PR #126, opened 2014-12-14
[EXPERIMENTAL] Add option :as for Mapper

### LINKED ISSUE #120: Remove Entity attributes inheritance
This is just an exploratory discussion.

I'm thinking if we want to allow Entity's inheritance.
Entities are Ruby objects that can be persisted by a Repository.

Imagine the following mapping:

``` ruby
collection :books do
  entity Book
  repository BookRepository

  attribute :id, Integer
  attribute :title, String
  attribute :price, Integer
end
```

Also imagine to have the following entities:

``` ruby
class Book
  include Lotus::Entity
  self.attributes = :title
end

class PricedBook < Book
  include Lotus::Entity
  self.attributes = :price
end
```

Now, this scenario may lead to some problems:
1. `Book` instances can't be fetched, because the mapping will try to invoke a setter for `price`, which isn't defined.
2. If I persist a `PricedBook` instance, when I do `BookRepository.find(23)` it will return a `Book`, because that is the entity mapped for that collection. This makes sense, but it's counter intuitive at the first glance.
3. This resembles like ActiveRecord Single Table Inheritance, which has been source of pain in my career.

We can proceed with two alternatives:
1. Prevent inheritance
2. Document that inheritance should be managed by mapping multiple times the same database table. To achieve this, we should introduce an aliasing for `Mapper#collection`. Right now it accepts a name that MUST match the table, and it also acts as unique identifier for the Mapper. We could to `collection :priced_books, as: :books`. Where the first will be the identifier and the second one the real table name.

What do you think?

joneslee85:
This override the table_name

Resolves #120 

The idea is to allow collection to have custom table name, this comes useful for the inherited entity usecase. Please see #120 for more details.

### DIFF
--- a/lib/lotus/model/adapters/sql_adapter.rb
+++ b/lib/lotus/model/adapters/sql_adapter.rb
@@ -146,7 +146,22 @@ module Lotus
         # @api private
         # @since 0.1.0
         def _collection(name)
-          Sql::Collection.new(@connection[name], _mapped_collection(name))
+          Sql::Collection.new(_dataset(name), _mapped_collection(name))
+        end
+
+        # Returns the dataset for Sql::Collection
+        #
+        # @param name [Symbol] a name of the collection
+        #
+        # @return [Sequel::Dataset]
+        #
+        # @see Lotus::Model::Adapters::Sql::Collection
+        #
+        # @api private
+        # @since 0.2.0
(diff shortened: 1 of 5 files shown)


joneslee85:
I close this PR for now

PR #137, opened 2014-12-31
Refactor attributes DSL

joneslee85:
* [x] Consolidate constructor
* [x] Gracefully handle unknown attributes

### DIFF
--- a/test/entity_test.rb
+++ b/test/entity_test.rb
@@ -87,15 +87,14 @@ describe Lotus::Entity do
         book.instance_variable_get(:@coolness).must_equal 'awesome'
       end
 
-      # TODO: this test does not test the correct behaviour - need to get it fixed
-      # it "doesn't interfer with superclass attributes" do
-      #   book = CoolNonFictionBook.new(title: "Good Math", author: "Mark C. Chu-Carroll", published: false, coolness: true)
+      it "doesn't interfer with superclass attributes" do
+        book = CoolNonFictionBook.new(title: "Good Math", author: "Mark C. Chu-Carroll", published: false, coolness: true)
 
-      #   book.instance_variable_get(:@title).must_equal  'Good Math'
-      #   book.instance_variable_get(:@author).must_equal 'Mark C. Chu-Carroll'
-      #   book.instance_variable_get(:@published).must_equal false
-      #   book.instance_variable_get(:@coolness).must_equal true
-      # end
+        book.instance_variable_get(:@title).must_equal  'Good Math'
+        book.instance_variable_get(:@author).must_equal 'Mark C. Chu-Carroll'
+        book.instance_variable_get(:@published).must_equal false
+        book.instance_variable_get(:@coolness).must_equal true
+      end
     end
 
     describe 'with undefined attributes' do
--- a/test/entity_test.rb
+++ b/test/entity_test.rb
@@ -82,14 +82,15 @@ describe Lotus::Entity do
         book.instance_variable_get(:@coolness).must_equal 'awesome'
       end
 
-      it "doesn't interfer with superclass attributes" do
-        book = CoolNonFictionBook.new(title: "Good Math", author: "Mark C. Chu-Carroll", published: false, coolness: true)
+      # TODO: this test does not test the correct behaviour - need to get it fixed
+      # it "doesn't interfer with superclass attributes" do
+      #   book = CoolNonFictionBook.new(title: "Good Math", author: "Mark C. Chu-Carroll", published: false, coolness: true)
 
-        book.instance_variable_get(:@title).must_equal  'Good Math'
-        book.instance_variable_get(:@author).must_equal 'Mark C. Chu-Carroll'
-        book.instance_variable_get(:@published).must_equal false
-        book.instance_variable_get(:@coolness).must_equal true
-      end
+      #   book.instance_variable_get(:@title).must_equal  'Good Math'
+      #   book.instance_variable_get(:@author).must_equal 'Mark C. Chu-Carroll'
+      #   book.instance_variable_get(:@published).must_equal false
+      #   book.instance_variable_get(:@coolness).must_equal true
+      # end
     end
 
     describe 'with undefined attributes' do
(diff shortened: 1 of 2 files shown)


joneslee85:
@AlfonsoUceda not yet, I have yet heard back from @jodosha, need more discussion

jodosha:
@joneslee85 @AlfonsoUceda That metaprogramming was introduced because of perf reasons.

### Metaprogramming vs Iterations (with only symbols)

At the beginning we were only supporting symbols as keys for the given set of attributes. If you have a look at this [bench](https://gist.github.com/jodosha/5c769a5e2f7b27c8b846#file-entity_initialize_symbol-rb) the results are impressive:

```
Calculating -------------------------------------
  symbol #initialize    42.997k i/100ms
    iter #initialize     9.567k i/100ms
-------------------------------------------------
  symbol #initialize    736.743k (±11.5%) i/s -      3.655M
    iter #initialize    113.413k (± 7.8%) i/s -    564.453k
```

The first implementation replicates the result of the metaprogramming that we want to remove with this PR. The second one represents the fallback that we want to keep. The results speaks for themselves: we should keep it because it's very, very fast.

### Metaprogramming vs Iterations (with indifferent access)

**However**, we decided to introduce indifferent access (via `Utils::Attributes#get`) because we expect external inputs to have both symbols and strings as keys for `attributes`. I suspect that `Utils::Attributes` is slow, because it [drags down the perf](https://gist.github.com/jodosha/5c769a5e2f7b27c8b846#file-entity_initialize_indifferent_access-rb):

```
Calculating -------------------------------------
    meta #initialize     7.437k i/100ms
    iter #initialize    11.623k i/100ms
-------------------------------------------------
    meta #initialize     84.850k (± 5.7%) i/s -    423.909k
    iter #initialize    137.717k (± 6.9%) i/s -    685.757k
```

According to this benchmark, we should remove metaprogramming.

### What to do?

One solution could be to symbolize the incoming attributes. However, this is only safe on MRI 2.2+ for now and it doesn't benefit from a huge [perf boost](https://gist.github.com/jodosha/5c769a5e2f7b27c8b846#file-entity_initialize_symbolized-rb):

```
Calculating -------------------------------------
symbolized #initialize
                        10.291k i/100ms
    iter #initialize     9.574k i/100ms
-------------------------------------------------
symbolized #initialize
                        116.616k (± 5.3%) i/s -    586.587k
    iter #initialize    108.023k (±12.4%) i/s -    536.144k
```

Metaprogramming pros:

  * Faster when attributes are symbols

Metaprogramming cons:

  * Slower for indifferent access
  * It's metaprogramming ;)
  * Attributes symbolize is only safe on MRI 2.2+

Iterative pros:

  * It doesn't require metaprogramming
  * Performs better with indifferent access
  * It's portable on all Ruby implementations

Iterative cons:

  * It's a bit slower when attributes are already passed as a symbolized hash

**Please remove metaprogramming** ;)

test/entity_test.rb
```diff
@@ -63,12 +63,6 @@ class Camera
         book.instance_variable_get(:@published).must_equal false
       end
 
-      it 'ignores unknown attributes' do
```
jodosha: @joneslee85 can you please get back this test?

We can guard for unknown attributes with:

```ruby
    attributes.each do |k, v|
      setter = "#{ k }="
      public_send(setter, v) if respond_to?(setter)
    end
  end
```

test/entity_test.rb
```diff
@@ -88,15 +82,15 @@ class Camera
         book.instance_variable_get(:@coolness).must_equal 'awesome'
       end
 
-      it "doesn't interfer with superclass attributes" do
-        book = Book.new(title: "Good Math", author: "Mark C. Chu-Carroll", published: false, price: 34, coolness: true)
+      # TODO: this test does not test the correct behaviour - need to get it fixed
```
jodosha: @joneslee85 I don't know what's wrong with this, but can you please fix it ;)

AlfonsoUceda:
nice explanation @jodosha

jodosha:
@joneslee85 Why you won't "gracefully handle unknown attributes"?

test/entity_test.rb
joneslee85: @jodosha I am not a fan of graceful unknown handling. Is there any pros of keeping this behaviour?

joneslee85:
@jodosha the metaprogramming constructor does handle this gracefully whilst the default constructor does not. I decided to opt for the ungraceful version, it is my solely my own opinion on it unless there are any advantages that I am not aware.

jodosha:
@joneslee85 Gotcha. Until now, the "default constructor" was a fallback for entities devs they don't declare attributes.

On before it was:

```ruby
class Book
  include Lotus::Entity
end

Book.new(title: 'The Tangled Web') # => ArgumentError: wrong number of arguments (1 for 0)
```

This were an unpleasant experience, that's why I introduced concrete `#initialize`.

Now, that we remove the one generated with metaprogramming (:+1:), that one will become the only one.

If now `Entity` constructor isn't graceful, it will force calling code to **inline rescue** or to **whitelist params**.

```ruby
book = Book.new(params) rescue nil

unless book.nil?
  # do something here
end
```

or

```ruby
attributes = ['title']
params = params.each_with_object({}) |(key, value), result|
   result[key] = value if attributes.include?(key.to_s)
end

book = Book.new(params)
# do something here
```

I'm not trying to reproduce mass assignment protection here, just to ignore unknown attributes in a safe way.

Exceptions are memory intensive. An attacker can send a lot of unknown attributes and exacerbate the server memory.

What do you think?

joneslee85:
@jodosha very good point on security though I think the concern you mention should be handle by Form Object or something at upper level. Atm I have no strong objecting opinion on this so I go with your suggestion.

jodosha:
@joneslee85 I agree with you on this:

> I think the concern you mention should be handle by Form Object or something at upper level

But don't forget that Model could be used independently from Lotus. We can't tell if other applications/frameworks will have those kind of protection. So having that guard over there has a lot of advantages.

jodosha:
@joneslee85 :+1:

PR #139, opened 2015-01-11
Added support for transactions

### LINKED ISSUE #133: Add SQL transactions support
This should be implemented in `Sql::Query`.
Here's the related Sequel documentation: http://sequel.jeremyevans.net/rdoc/files/doc/transactions_rdoc.html

I'd say to not support/document `Sequel::Rollback`. Using exceptions for control flow is always a poor choice. If an exception is raised as side effect, the transaction should be rolled back.

```ruby
transaction do
  raise `Sequel::Rollback` # this will rollback anyway because of Sequel, but it's a poor style that we shouldn't encourage
end
```

```ruby
transaction do
  some_dangerous_method! # if this raises an exception, the transaction should be rolled back
end
```

The second style is the one that we should encourage/document.

/cc @stevehodgkiss @joneslee85

AlfonsoUceda:
Closes #133

### DIFF
--- a/lib/lotus/model/adapters/abstract.rb
+++ b/lib/lotus/model/adapters/abstract.rb
@@ -162,6 +162,18 @@ module Lotus
         def query(collection, &blk)
           raise NotImplementedError
         end
+
+        # Create a transaction
+        # Only supported by SqlAdapter
+        #
+        # @param options [Hash] options for transaction
+        #
+        # @see Lotus::Model::Adapters::SqlAdapter#transaction
+        #
+        # @since x.x.x
+        def transaction(options = {})
+          yield
+        end
       end
     end
   end
--- a/test/repository_test.rb
+++ b/test/repository_test.rb
@@ -299,4 +299,77 @@ describe Lotus::Repository do
       end
     end
   end
+
+  describe "with sql adapter" do
+    before do
+      UserRepository.adapter    = Lotus::Model::Adapters::SqlAdapter.new(MAPPER, SQLITE_CONNECTION_STRING)
+      ArticleRepository.adapter = Lotus::Model::Adapters::SqlAdapter.new(MAPPER, SQLITE_CONNECTION_STRING)
+
+      UserRepository.collection    = :users
+      ArticleRepository.collection = :articles
+
+      UserRepository.clear
+      ArticleRepository.clear
+
+      ArticleRepository.create(article1)
+    end
+
+    describe '.transaction' do
+      it 'if an exception is raised the size of articles is equal to 1' do
+        ArticleRepository.all.size.must_equal 1
+        exception = -> { ArticleRepository.transaction do
+          ArticleRepository.create(article2)
+          raise Exception
+        end }
+        exception.must_raise Exception
+        ArticleRepository.all.size.must_equal 1
+      end
+
+      it "if an exception isn't raised the size of articles is equal to 2" do
+        ArticleRepository.all.size.must_equal 1
+        ArticleRepository.transaction do
+          ArticleRepository.create(article2)
+        end
+        ArticleRepository.all.size.must_equal 2
+      end
+
+      describe 'using options' do
+        it 'rollback: :always option' do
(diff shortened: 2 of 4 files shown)


jodosha:
@AlfonsoUceda Thanks for this PR.

This looks good, but we need to test the type of transactions that we are going to support. Sequel has fairly advanced features for this. We're not helping enough developers to understand how to use those features and protecting them from changes.

### Help

I'm for expanding API docs with the options that we support and with examples.

### Changes

We depend on `sequel ~> 4.9`. If one of those options change in a minor version we cannot guarantee that Lotus::Model works with it.

```ruby
UserRepository.transaction(retry_on: [RecordNotUnique]) do
  UserRepository.create(User.new(username: 'jodosha'))
end
```

Now suppose that a new version of Sequel renames that `:retry_on` with `:retry`, the applications will suddenly break without us to even notice. The reason is we don't have tests to cover those cases.

The result is a breaking change for developers even if we haven't bumped to a version that allows us to do so. In other words: they depend on `lotus-model` which should guarantee that the application will work until they don't upgrade to a new version of it.

### Adapters interoperability

In order to increase the compatibility between adapters, I'm for removing that exception on the memory adapter and to document this limitation. The implementation should just yield the given block and do nothing more.



Can you please expand this PR? Thank you very much.

AlfonsoUceda:
If I undestood well, we have to test every option that sequel 4.9 support in transactions to ensure breaks if sequel is updated.

In the `Adapters interoperability` section, I comment in code.

lib/lotus/model/adapters/abstract.rb
```diff
@@ -162,6 +170,15 @@ def command(query)
...
           raise NotImplementedError
         end
+
+        # Create a transaction
+        #
+        # @param options [Hash] options for transaction
+        #
+        # @since x.x.x
+        def transaction(options = {})
+          raise MethodNotSupported
```
AlfonsoUceda: remove this Exception and only yield? and document that only available in sql adapter?

test/repository_test.rb
```diff
@@ -299,4 +299,61 @@ def to_int
...
+        ArticleRepository.all.size.must_equal 1
+        ArticleRepository.transaction do
+          ArticleRepository.create(article2)
+        end
+        ArticleRepository.all.size.must_equal 2
+      end
+    end
+  end
+
+  describe "with memory adapter" do
```
AlfonsoUceda: don't test this case, isn't it?

jodosha:
@AlfonsoUceda Yes. My point is: we should test the options that we want to support. I'd say we can start with the following list for the next release:

* `rollback: :reraise`
* `rollback: :always`

So the API docs and the tests should be explicit on those cases. That's it :)

test/repository_test.rb
jodosha: I'd say to keep the test to document the existence of `MemoryAdapter#transaction`, but the test should not to reproduce SQL behavior (rollback), only assert what it does.

```ruby
-> {
BookRepository.transaction do
  raise "boom"
end
}.must_raise RuntimeError
```

;)

lib/lotus/model/adapters/abstract.rb
jodosha: :+1:

AlfonsoUceda:
@jodosha what do you think about these changes?

jodosha:
@AlfonsoUceda This looks good, I'll do a minor API docs cleanup. Thank you! :+1:

AlfonsoUceda:
You welcome :D

jodosha:
@joneslee85 Can you please review this when you have time? TY!

test/repository_test.rb
joneslee85: From user perspective, I think it's worth to raise something a meaningful message that the adapter does not support transaction rather than a generic RuntimeError. But that could be another PR.

joneslee85:
Me happy :+1:

jodosha:
@AlfonsoUceda Thank you, I've just merged this. :+1:

PR #141, opened 2015-01-13
Coerce entities when persisted

### LINKED ISSUE #138: Attributes not coerced when persisted
```ruby
class User
  include Lotus::Entity
  attributes :id, :name, :age
end

mapper = Lotus::Model::Mapper.new do
  collection :users do
    entity User

    attribute :id,   String
    attribute :name, String
    attribute :age,  Integer
  end
end.load!

user1 = User.new(name: 'Luca', age: '32')
UserRepository.persist(user1)
```

When `user1` is persisted, the age is **not** coerced to an `Integer` before persisting and remains as a string. For dynamic datastores (e.g. RethinkDB), the type is kept intact.

When retrieved from storage, `user1.age` is coerced to an `Integer` from a `String`.

Is this behavior intended? I can see how input should be type-validated before it reaches the persisting method, but defining the type in the mapper should guarantee the type is saved properly.

Is it the job of the adapter to coerce the attributes?

jodosha:
Coerce entities when persisted via repositories: `.persist`, `.create` and `.update`.
Those operations will return coerced entities, **without the need of reloading** the record from the database.

Entities are now coerced before to be stored. This will now help schema less databases to receive the expected values (#138).

```ruby
# BEFORE

user = User.new(name: 'L', age: '32')
UserRepository.create(user) # in place assignment of `id`

user.id   # => 1
user.name # => "L"
user.age  # => "32" # WRONG, not coerced

# AFTER

user = User.new(name: 'L', age: '32')
user = UserRepository.create(user) # returns a new instance of `User` with coerced values and `id`

user.id   # => 1
user.name # => "L"
user.age  # => 32 # RIGHT
```

/cc @lotus/core-team 

Closes #138

### DIFF
--- a/lib/lotus/model/mapping/coercer.rb
+++ b/lib/lotus/model/mapping/coercer.rb
@@ -58,13 +58,16 @@ module Lotus
           instance_eval %{
             def to_record(entity)
               if entity.id
-                Hash[#{ @collection.attributes.map{|name,(_,mapped)| ":#{mapped},entity.#{name}"}.join(',') }]
+                Hash[#{ @collection.attributes.map{|name,(klass,mapped)| ":#{mapped},Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name})"}.join(',') }]
               else
-                Hash[#{ @collection.attributes.reject{|name,_| name == @collection.identity }.map{|name,(_,mapped)| ":#{mapped},entity.#{name}"}.join(',') }]
+                Hash[#{ @collection.attributes.reject{|name,_| name == @collection.identity }.map{|name,(klass,mapped)| ":#{mapped},Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name})"}.join(',') }]
               end
             end
 
             def from_record(record)
+            if $debug == true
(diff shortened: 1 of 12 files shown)


joneslee85:
:sparkling_heart: I love it :+1:

lib/lotus/model/mapping/coercer.rb
```diff
@@ -58,13 +58,16 @@ def deserialize_#{ mapped }(value)
...
               else
-                Hash[#{ @collection.attributes.reject{|name,_| name == @collection.identity }.map{|name,(_,mapped)| ":#{mapped},entity.#{name}"}.join(',') }]
+                Hash[#{ @collection.attributes.reject{|name,_| name == @collection.identity }.map{|name,(klass,mapped)| ":#{mapped},Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name})"}.join(',') }]
               end
             end
 
             def from_record(record)
+            if $debug == true
+                raise Hash[#{ @collection.attributes.map{|name,(klass,mapped)| ":#{name},Lotus::Model::Mapping::Coercions.#{klass}(record[:#{mapped}])"}.join(',') }].inspect
+                end
```
stevehodgkiss: Did you intend to merge this?

jodosha: Fixed by af99b67 thank you! :100:

PR #142, opened 2015-01-18
Add file and line to instance_eval call to make backtrace more usable.

midas:
Avoid situations where the backtrace includes an (eval) instead of the name of
the file:

    from (eval):5:in `foo'
    from foo.rb:11

### DIFF
--- a/lib/lotus/model/mapping/coercer.rb
+++ b/lib/lotus/model/mapping/coercer.rb
@@ -55,7 +55,7 @@ module Lotus
             }
           end.join("\n")
 
-          instance_eval %{
+          instance_eval <<-EVAL, __FILE__, __LINE__
             def to_record(entity)
               if entity.id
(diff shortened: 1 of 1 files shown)


AlfonsoUceda:
that's is testable?

midas:
No test written for this as it is just changing the parameters sent to the instance_eval method call and should be covered by the exercising of the methods defined within the eval.  

Not sure why code coverage went up by 0.01%.  Kind of funny.  Possibly the implementation of the code coverage calculator benefitted from the __FILE__ and __LINE__ information in the backtrace to aid in its calculation?  I am probably stretching here.

joneslee85:
:+1:

jodosha:
@midas Thanks for this PR, this looks good :+1: 

Coveralls, it's not so accurate. We see some minor changes from PR to PR.

PR #144, opened 2015-01-25
Add Lotus::Model::Migration

### LINKED ISSUE #115: Add migration mechanism
**UPDATE**: Please see proposed changes in my PR https://github.com/lotus/model/issues/144

What: Schema Migration for SQL adapter
How:
* Wrap [Sequel::Migrator](http://sequel.jeremyevans.net/rdoc/classes/Sequel/Migrator.html) vs Writing from scratch?
* Migrator class has `up`, `down` method
* Schema file is the source of truths, and it should not reflect local changes of DB schema.
* There is `db/schema.rb` for `/lib` and `apps/<app_name>/db/schema.rb` for app
* Stop users from abusing migration to do data migration (I know - the wording 'migration' is confusing eh?)
* Allow migrations to be stored in 1 level deep or many level deep folder under (as like Rails)
* Follow Sequel Migrator DSL vs Rails DSL
* Generator is to be done in lotusrb gem, and the core classes is to be done in model gem
* The UX for the generator:

  + `lotus generate migration oh_my_god` will generate migration for `/lib`
  + `LOTUS_APP=web lotus generate migration oh_my_lord` will generate migration for `/apps/web`
  OR
  + `lotus generate migration web:oh_my_lord`

joneslee85:
[Fixes #115]

This adds the basic migration feature for Lotus.

Please be noted that this PR does not cover load/dump schema.rb, I am going to have a separate PR for that.

## TODO

- [X] Migration class
- [X] Timestamp Migrator (Noted: this is totally different to Sequel::TimestampMigrator)
- [ ] Schema.rb dump
- [ ] Write CLI task 'lotus generate migration' to generate new migration
- [ ] Write CLI task 'lotus db migrate' to migrate
- [ ] Write CLI task 'lotus db rollback' to rollback
- [ ] Write CLI task 'lotus db apply' to merge all changes to schema.rb and remove migration files

## Migration

It purely base on Sequel::Migration and thus share the same interface as Sequel::Migration:

```ruby
class MyMigration < Lotus::Model::Migration
  def up
  end

  def down
  end
end
```

Migration files are located inside `LOTUS_ROOT/db/migrations` folder by default. However you can configure it with `Lotus::Model::Configuration#migration_directory`

## Migrator

To migrate, we use Migrator:

```ruby
require 'lotus/model'
require 'lotus/model/migrator'

#adapter configuration is copied from the framework configuration
# so make sure you have to configure framework first or else
# you will get error!    
Lotus::Model.configure do
  migration_directory 'my_apps/db/migrations'
  adapter type: :sql, uri: 'sqlite3://localhost/database'
end
    
Lotus::Model::Migrator.new.run
```

To migrate from a directory

```ruby
Lotus::Model::Migrator.new.migrate
```

To rollback to previous migration

```ruby
Lotus::Model::Migrator.new.rollback
```

or migration 4 steps from the last version

```ruby
Lotus::Model::Migrator.new.rollback(step: 4)
```

Furthermore, I also introduce the ability to set logger for the framework. Please see `Lotus::Model::Configuration#logging`. The migrator will use the framework logger implicitly.

### create_table

Use to create new table. Syntax to add new column is below:

```ruby
create_table(:songs) do
  column :id, primary_key: true
  column :title, :string, null: false
  column :artist, :string, 'VARCHAR(15)'
end
```

### DIFF
--- a/lib/lotus/model.rb
+++ b/lib/lotus/model.rb
@@ -72,8 +72,9 @@ module Lotus
     #     end
     #   end
     #
-    # Adapter MUST follow the convention in which adapter class is inflection of adapter name
-    # The above example has name :sql, thus derived class will be `Lotus::Model::Adapters::SqlAdapter`
+    # Adapter MUST follow the convention in which adapter class is inflection
+    # of adapter name. The above example has name :sql, thus derived class will
+    # be `Lotus::Model::Adapters::SqlAdapter`.
     def self.configure(&blk)
       configuration.instance_eval(&blk)
       self
@@ -184,7 +185,9 @@ module Lotus
     #   end
     #
     #   Lotus::Model.configuration.adapter_config # => nil
-    #   MyApp::Model.configuration.adapter_config # => #<Lotus::Model::Config::Adapter:0x007ff0ff0244f8 @type=:memory, @uri="memory://localhost", @class_name="MemoryAdapter">
+    #   MyApp::Model.configuration.adapter_config
+    #     => #<Lotus::Model::Config::Adapter:0x007ff0ff0244f8 @type=:memory,
+    #          @uri="memory://localhost", @class_name="MemoryAdapter">
     def self.duplicate(mod, &blk)
       dupe.tap do |duplicated|
         mod.module_eval %{
--- a/lib/lotus/model/adapters/sql/schema_modifier.rb
+++ b/lib/lotus/model/adapters/sql/schema_modifier.rb
@@ -64,7 +64,7 @@ module Lotus
         # @api private
         class SchemaModifier
           def initialize(adapter, logger = nil)
-            @connection = adapter.instance_variable_get(:@connection)
+            @connection = adapter.connection
             @connection.logger = logger if logger
           end
 
--- a/lib/lotus/model/adapters/sql/schema_modifier.rb
+++ b/lib/lotus/model/adapters/sql/schema_modifier.rb
@@ -1,0 +1,99 @@
+require 'sequel'
+
+module Lotus
+  module Model
+    module Adapters
+      module Sql
+        # Allow users to easily group schema changes and migrate the
+        # database to a newer version or revert to a previous version.
+        #
+        # A migration class must be subclass of Lotus::Model::Migration
+        # and must have 2 instance methods:
+        #
+        #   * up   - to migrate
+        #   * down - to rollback
+        #
+        # Both instance methods could access following schema modification methods:
+        #
+        #   * add_column
+        #   * add_index
+        #   * create_view
+        #   * drop_column
+        #   * drop_index
+        #   * drop_table
+        #   * drop_view
+        #   * rename_table
+        #   * rename_column
+        #   * set_column_default
+        #   * set_column_type
+        #
+        # For more details, please consult documentation
+        # at http://sequel.jeremyevans.net/rdoc/files/doc/schema_modification_rdoc.html
+        #
+        # To apply up/down migration, please consult Lotus::Model::Migration#run
+        # Please avoid running running migration through the low level
+        # Sequel API because doing so does not update the schema_versions table
+        # and thus will create inconsistency in migration history.
+        #
+        # Due to the fact the Sequel is going to kill old class based migration,
+        # it is decided to clone source code from Sequel::Migration.
+        #
+        # @example Create and apply migration
+        #   require 'lotus/model/migration'
+        #
+        #   class MyMigration < Lotus::Model::Migration
+        #     def up
+        #       create_table(:authors) do
+        #         primary_key :id
+        #         String :name
+        #       end
+        #     end
+        #
+        #     def down
+        #       drop_table(:authors)
+        #     end
+        #   end
+        #
+        #   # to apply up migration
+        #   MyMigration.new(adapter).up
+        #
+        #   # to apply down migration
+        #   MyMigration.new(adapter).down
+        #
+        # @since x.x.x
+        # @api private
+        class SchemaModifier
+          def initialize(adapter, logger = nil)
+            @connection = adapter.instance_variable_get(:@connection)
+            @connection.logger = logger if logger
+          end
+
+          private
+
+          # Intercepts method calls intended for the database and sends them along.
+          #
+          # @since x.x.x
+          # @api private
+          def method_missing(method_sym, *args, &block)
+            connection.send(method_sym, *args, &block)
+          end
+
+          # This object responds to all methods the database responds to.
+          #
+          # @since x.x.x
+          # @api private
+          def respond_to_missing?(meth, include_private)
+            connection.respond_to?(meth, include_private)
+          end
+
+          # The DB Connection
+          #
+          # @since x.x.x
+          # @api private
+          attr_reader :connection
+
+        end
+      end
+    end
+  end
+end
--- a/lib/lotus/model/configuration.rb
+++ b/lib/lotus/model/configuration.rb
@@ -187,7 +187,7 @@ module Lotus
       #   Lotus::Model.configuration.migrations_directory
       #     # => 'my_custom/migrations'
       #
-      # @since 0.3.0
+      # @since x.x.x
       # @see Lotus::Model.configure
       def migrations_directory(directory=nil)
         if directory.nil?
@@ -283,7 +283,7 @@ module Lotus
 
       # Default directory that contains migration files
       #
-      # @since 0.3.0
+      # @since x.x.x
       # @api private
       def _default_migrations_directory
         Pathname.pwd.join('db', 'migrations')
--- a/lib/lotus/model/configuration.rb
+++ b/lib/lotus/model/configuration.rb
@@ -1,5 +1,6 @@
 require 'lotus/model/config/adapter'
 require 'lotus/model/config/mapper'
+require 'logger'
 
 module Lotus
   module Model
@@ -56,9 +57,17 @@ module Lotus
       # @since 0.2.0
       def load!
         _build_mapper
-        _build_adapter
+        build_adapter
 
         mapper.load!(@adapter)
+      end
+
+      # Instantiate adapter from adapter_config
+      #
+      # @api private
+      # @since 0.1.0
+      def build_adapter
+        @adapter = adapter_config.build(mapper)
       end
 
       # Register adapter
@@ -137,6 +146,98 @@ module Lotus
         @mapper_config = Lotus::Model::Config::Mapper.new(path, &blk)
       end
 
+      # Set the logger
+      #
+      # @param logger [Logger] the logger instance, default to ::Logger STDOUT
+      #
+      # @example Set custom logger and retrieve the set logger
+      #   require 'lotus/model'
+      #
+      #   Lotus::Model.configure do
+      #     logger ::Logger.new(STDOUT)
+      #   end
+      #
+      #   Lotus::Model.configuration.logger
+      #     # => #<Logger:0x007ffc45227920>
+      #
+      # @since 0.3.0
+      # @see Lotus::Model.configure
+      def logger(logger = nil)
+        if logger.nil?
+          @logger ||= _default_logger
+        else
+          @logger = logger
+        end
+      end
+
+      # Set directory where migration files should be migrated from
+      #
+      # @param directory [String] the migration directory filepath
+      #
+      # @example Set custom logger and retrieve the set logger
+      #   require 'lotus/model'
+      #
+      #   Lotus::Model.configuration.migration_directory
+      #     # => 'db/migrations'
+      #
+      #   Lotus::Model.configure do
+      #     migration_directory 'my_custom/migrations'
+      #   end
+      #
+      #   Lotus::Model.configuration.migration_directory
+      #     # => 'my_custom/migrations'
+      #
+      # @since 0.3.0
+      # @see Lotus::Model.configure
+      def migration_directory(directory=nil)
+        if directory.nil?
+          @migration_directory ||= _default_migration_directory
+        else
+          @migration_directory = directory
+        end
+      end
+
+      # Return a copy of the configuration of the framework instance associated
+      # with the given class.
+      #
+      # When multiple instances of Lotus::Model are used in the same
+      # application, we want to make sure that a model or a migrator will
+      # receive the expected configuration.
+      #
+      # @param base [Class, Module] a model or a migrator
+      #
+      # @return [Lotus::Model::Configuration] the configuration associated
+      #   to the given class.
+      #
+      # @since 0.3.0
+      # @api private
+      #
+      # @example Direct usage of the framework
+      #   require 'lotus/model'
+      #
+      #   Lotus::Model::Configuration.for(Migrator)
+      #     # => will duplicate from Lotus::Model
+      #
+      # @example Multiple instances of the framework
+      #   require 'lotus/model'
+      #
+      #   module MyApp
+      #     module Model
+      #       Migrator = ::Lotus::Model::Migrator.duplicate(self)
+      #     end
+      #   end
+      #
+      #   Lotus::Model::Configuration.for(Migrator)
+      #     # => will duplicate from Lotus::Model
+      #
+      #   Lotus::Model::Configuration.for(MyApp::Model::Migrator)
+      #     # => will duplicate from MyApp::Model
+      def self.for(base)
+        namespace = Utils::String.new(base).namespace
+        framework = Utils::Class.load_from_pattern!("(#{namespace}|Lotus)::Model")
+        framework.configuration.duplicate
+      end
+
       # Duplicate by copying the settings in a new instance.
       #
       # @return [Lotus::Model::Configuration] a copy of the configuration
@@ -146,7 +247,10 @@ module Lotus
       def duplicate
         Configuration.new.tap do |c|
           c.instance_variable_set(:@adapter_config, @adapter_config)
+          c.instance_variable_set(:@adapter, @adapter)
           c.instance_variable_set(:@mapper, @mapper)
+          c.instance_variable_set(:@logger, @logger)
+          c.instance_variable_set(:@migration_directory, @migration_directory)
         end
       end
 
@@ -163,12 +267,6 @@ module Lotus
       end
 
       # @api private
-      # @since 0.1.0
-      def _build_adapter
-        @adapter = adapter_config.build(mapper)
-      end
-
-      # @api private
       # @since 0.2.0
       #
       # NOTE Drop this manual check when Ruby 2.0 will not be supported anymore.
@@ -181,6 +279,22 @@ module Lotus
         [:type, :uri].each do |keyword|
           raise ArgumentError.new("missing keyword: #{keyword}") if !options.keys.include?(keyword)
         end
+      end
+
+      # Default directory that contains migration files
+      #
+      # @since 0.3.0
+      # @api private
+      def _default_migration_directory
+        Pathname.pwd.join('db', 'migrations')
+      end
+
+      # Default logger to stdlib logger
+      #
+      # @since 0.3.0
+      # @api private
+      def _default_logger
+        ::Logger.new(STDOUT)
       end
     end
   end
--- a/lib/lotus/model/configuration.rb
+++ b/lib/lotus/model/configuration.rb
@@ -177,23 +177,23 @@ module Lotus
       # @example Set custom logger and retrieve the set logger
       #   require 'lotus/model'
       #
-      #   Lotus::Model.configuration.migration_directory
+      #   Lotus::Model.configuration.migrations_directory
       #     # => 'db/migrations'
       #
       #   Lotus::Model.configure do
-      #     migration_directory 'my_custom/migrations'
+      #     migrations_directory 'my_custom/migrations'
       #   end
       #
-      #   Lotus::Model.configuration.migration_directory
+      #   Lotus::Model.configuration.migrations_directory
       #     # => 'my_custom/migrations'
       #
       # @since 0.3.0
       # @see Lotus::Model.configure
-      def migration_directory(directory=nil)
+      def migrations_directory(directory=nil)
         if directory.nil?
-          @migration_directory ||= _default_migration_directory
+          @migrations_directory ||= _default_migrations_directory
         else
-          @migration_directory = directory
+          @migrations_directory = directory
         end
       end
 
@@ -250,7 +250,7 @@ module Lotus
           c.instance_variable_set(:@adapter, @adapter)
           c.instance_variable_set(:@mapper, @mapper)
           c.instance_variable_set(:@logger, @logger)
-          c.instance_variable_set(:@migration_directory, @migration_directory)
+          c.instance_variable_set(:@migrations_directory, @migrations_directory)
         end
       end
 
@@ -285,7 +285,7 @@ module Lotus
       #
       # @since 0.3.0
       # @api private
-      def _default_migration_directory
+      def _default_migrations_directory
         Pathname.pwd.join('db', 'migrations')
       end
 
--- a/lib/lotus/model/migration.rb
+++ b/lib/lotus/model/migration.rb
@@ -1,0 +1,128 @@
+require 'delegate'
+require 'lotus/utils/class_attribute'
+require 'lotus/model/adapters/sql/schema_modifier'
+
+module Lotus
+  module Model
+    # Allow users to easily group schema changes and migrate the
+    # database to a newer version or revert to a previous version.
+    #
+    # A migration class must be subclass of Lotus::Model::Migration
+    # and must implement 2 instance methods:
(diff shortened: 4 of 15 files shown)


lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,90 @@
+require 'sequel'
+require 'sequel/extensions/migration'
+
+# Allow users to easily group schema changes # and migrate the
```
AlfonsoUceda: `changes # and`
to
`changes and`

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,90 @@
...
+#   adapter_config = Lotus::Model::Config::Adapter.new(type: :sql, uri: DB_URI)
+#   # to migrate
+#   MyMigration.apply(adapter_config, :up)
+#
+#   # to rollback
+#   MyMigration.apply(adapter_config, :down)
+#
+# @since x.x.x
+# @see Sequel::Migration
+class Lotus::Model::Migration < Sequel::Migration
```
AlfonsoUceda: instead that, what do you think with this?
```ruby
module Lotus
  module Model
    class Lotus::Model::Migration < Sequel::Migration
    end
  end
end
```

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,90 @@
...
+  # direction.
+  #
+  # @param adapter_config - instance of Lotus::Model::Config::Adapter
+  #
+  # @since x.x.x
+  def self.apply(adapter_config, direction)
+    raise(ArgumentError, "Invalid migration direction specified (#{direction.inspect})") unless [:up, :down].include?(direction)
+    new(adapter_config).send(direction)
+  end
+
```
AlfonsoUceda: remove space

AlfonsoUceda:
@joneslee85 have you though to include a step option to rollback? like rails does: `rake db:migrate STEP=2` and you rollback 2 migrations

lib/lotus/model/migration.rb
joneslee85: ? Do you mean?

```
module Lotus
  module Model
    class Migration < Sequel::Migration
    end
  end
end
```

joneslee85:
@AlfonsoUceda I have yet touched the migrator part, but it will be coming as separate PR. This PR is only for the core part.

lib/lotus/model/migration.rb
AlfonsoUceda: ups, yes sorry

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,163 @@
...
+      # Run the migrator
+      #
+      # @since x.x.x
+      # @delegate Sequel::TimestampMigrator#run
+      #
+      # @see Sequel::TimestampMigrator#run
+      delegate :run => :migrator
+
+      MIGRATION_DIRECTORY = "#{Dir.pwd}/db/migrations".freeze
```
jc00ke: Do we always use `Dir.pwd` or is there a notion of `Lotus.root`?

joneslee85: Come to think of it, there is `Lotus::Environment`, I am going to see if I could use it

joneslee85: @jc00ke Because lotus/model is going to be independent of lotusrb, thus I am going to stick with the current path.

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,157 @@
+require 'sequel'
+require 'sequel/extensions/migration'
+require 'forwardable'
+require 'logger'
+
+module Lotus
+  module Model
+    # Allow users to easily group schema changes # and migrate the
```
AlfonsoUceda: /s/changes # and/ changes and/

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,157 @@
...
+        @migrator = ::Sequel::TimestampMigrator.new(db, directory, {target: target, current: current})
+        _set_logger(logger)
+      end
+
+      # A shorthand to invoke +run+
+      #
+      # @since x.x.x
+      #
+      # @see Sequel::Migrator#initialize
+      def self.apply(db, **options)
```
AlfonsoUceda: comment the params

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,157 @@
...
+      #
+      # @return [Lotus::Model::Migrator] a new migrator instance
+      #
+      # @since x.x.x
+      #
+      # @see Sequel::TimestampMigrator
+      def initialize(adapter_config, directory: MIGRATION_DIRECTORY, target: nil, current: nil, logger: ::Logger.new(STDOUT))
+        raise Lotus::Model::Migrator::UnsupportedAdapter if adapter_config.type != :sql
+        @db = ::Sequel.connect(adapter_config.uri)
+        @migrator = ::Sequel::TimestampMigrator.new(db, directory, {target: target, current: current})
```
AlfonsoUceda: /s/{target: target, current: current}/{ target: target, current: current }

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,157 @@
...
+    #
+    #   # to migrate
+    #   MyMigration.apply(DB, :up)
+    #
+    #   # to rollback
+    #   MyMigration.apply(DB, :down)
+    #
+    # @since x.x.x
+    # @see Sequel::Migration
+    Migration = ::Sequel::Migration
```
AlfonsoUceda: Is this used?

AlfonsoUceda:
:+1: resolve comments and test rollback

lib/lotus/model/migration.rb
joneslee85: I am going to stick without spaces because it is now a convention that we follow ;)

AlfonsoUceda: ok ;)

AlfonsoUceda:
looks good for me :)

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,161 @@
...
+    #   * drop_view
+    #   * rename_table
+    #   * rename_column
+    #   * set_column_default
+    #   * set_column_type
+    #
+    # For more details, please consult documentation
+    # at http://sequel.jeremyevans.net/rdoc/files/doc/schema_modification_rdoc.html
+    #
+    # @xample
```
7even: I believe this should be `@example`.

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,161 @@
...
+
+      private
+
+      # Set logging to +Lotus::Logger+
+      #
+      # @since x.x.x
+      # @api private
+      def _set_logger(logger)
+        db.loggers << logger
+      end
```
huydx: what are merits to keep multiple loggers in migration?

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,161 @@
...
+      # @param target [String] the target migration file
+      # @param current [String] the current migration version
+      #
+      # @return [Lotus::Model::Migrator] a new migrator instance
+      #
+      # @since x.x.x
+      #
+      # @see Sequel::TimestampMigrator
+      def initialize(adapter_config, directory: MIGRATION_DIRECTORY, target: nil, current: nil, logger: ::Logger.new(STDOUT))
+        raise Lotus::Model::Migrator::UnsupportedAdapter if adapter_config.type != :sql
```
huydx: [nitpick] i think the dependency of supported type should be define elsewhere, should not be hard-coded like this ;)

janlelis:
Hi everybody,
I want to suggest a small detail about migrations: Some years ago I wrote a mini library for Rails migrations that would do no more than replacing
```ruby
class SomeMigration < ActiveRecord::Migration
  # ...
```
with 
```ruby
migration do
  # ...
```
which generated the class name automatically, based on the filename. The motivation behind this was that when I wanted to create a new migration (without using Rails' generator), I would always need to come up with a new, unique class name, that I never needed again. Having a short `migration do` solved this problem for me.

Unfortunately, it introduces magic, but something similar/better could be done in a not-so-magic manner:

```ruby
Lotus::Model::Migration.create "description" do
  # ... (the description being also generated, but it is an optional parameter)
```

I propose this little API change, because:
- It frees you from making up new constant names
- Instead, it gives you the possibility to describe what you want to do in a free format (rspec describe style)
- The description string can be used to generate an unique migration class (random one, if none given)

I am happy to write the code for this, if you like the idea.

joneslee85:
@janlelis Thanks much for your detailed feedback.

I have been looking into the DSL solution from the very start of this feature. In fact, doing DSL is very straight-forward because Sequel already had DSL for migration.

After weighting pros/cons, I decided to stay away from DSL to avoid unnecessary complexity.

I acknowledge your pain point regarding unique class constant however I don't think it would be a huge hassle because Lotus would provide migration generator. Besides, your suggestion on class name inflection from filename is not going to work out of the box though, for schema_migration table stores the full <version>_filepath into filename column and should users have migrated (not rollback yet) then rename the filename would create extra record in schema_migration table.

Having said all of the above, I am open for feedbacks from other uses, if this is truly a pain point, I am more than happy to make changes.

mwallasch:
@joneslee85 Great work. One small question: is there a good reason to keep the current_version method private. Sometimes it would be helpful to have the possibility for a quick check for the current migrated version.

joneslee85:
@mwallasch there is no particular reason. I did not think of your scenario. Thanks for the feedback. I'll make change.

jodosha:
@joneslee85 In the description and in the inline docs you say that the default logger is `Lotus::Logger`, which is included by `lotusrb`, but in the end the code uses Ruby's stdlib `::Logger`. This because it makes it hard to reuse that component from our main gem. Do you think we should extract it into `lotus-utils`, so it can be available everywhere?

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,277 @@
...
+    #       end
+    #     end
+    #
+    #     def down
+    #       drop_table(:authors)
+    #     end
+    #   end
+    #
+    #   # to migrate
+    #   MyMigration.apply(DB, :up)
```
jodosha: @joneslee85 This comment puzzles me. :smiley_cat: `DB` derives from Sequel, but what's the meaning here in Lotus::Model?

joneslee85:
@jodosha I forget to update the description. I don't think it is necessary to extract it to lotus/utils. My plan is to support `::Logger` in lotus/model and leave it to the migration CLI of lotus/lotus to set logger to `Lotus::Logger`.

lib/lotus/model/migration.rb
joneslee85: Yes, the `Migration` class is low-level class which has access to `apply` method and that method takes in a sequel connection. I think I am going to document that for whom who care about this low level detail or should I remove it?

lib/lotus/model/migration.rb
```diff
@@ -0,0 +1,277 @@
...
+      # It's raised when migration does not support the adapter
+      #
+      # @since 0.3.0
+      class UnsupportedAdapter < ::StandardError
+        def initialize(adapter_type)
+          super("Adapter #{adapter_type} is not supported")
+        end
+      end
+
+      MIGRATION_DIRECTORY = "#{Dir.pwd}/db/migrations".freeze
```
jodosha: Should we split the composition of this directory? If feel uneasy to leave a frozen constant with the current value that can possibly change. Eg. if someone uses `Dir.chdir`. Can we interpolate at the run time?

Example:

```
MIGRATION_DIRECTORY = 'db/migrations'.freeze
```

And then in the methods that use it:

```
directory = "#{ Dir.pwd }/MIGRATION_DIRECTORY"
```

We can have a private method to do this tedious interpolation over and over.

lib/lotus/model/migration.rb
jodosha: Well, we need to document anyway how to apply a migration. Please keep the documentation, but with a Lotus terminology. :smile:

lib/lotus/model.rb
```diff
@@ -187,3 +187,6 @@ def self.duplicate(mod, &blk)
 
   end
 end
+
+require 'lotus/model/migration'
```
AlfonsoUceda: why at the end?

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,231 @@
...
+    # stored in the database, the version is considered to be 0.
+    # if no target version is specified, the database is migrated
+    # to the latest version available in migration directory.
+    #
+    # @example Migrate all migrations in db/migrations to latest
+    #   require 'lotus/model/migration'
+    #   adapter_cfg = Lotus::Model::Config::Adapter.new(type: :sql, uri: DATABASE_URI)
+    #   Lotus::Model::Migrator.new(adapter_cfg).run
+    #
+    # @example To use shortcut +apply+ method to invoke +run+ instance method
```
AlfonsoUceda: this is the same example like above

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,231 @@
...
+      # It's raised when migration does not support the adapter
+      #
+      # @since 0.3.0
+      class UnsupportedAdapter < ::StandardError
+        def initialize(adapter_type)
+          super("Adapter #{adapter_type} is not supported")
+        end
+      end
+
+      MIGRATION_DIRECTORY = '/db/migrations'.freeze
```
AlfonsoUceda: these constants should be commented

AlfonsoUceda:
:+1:

lib/lotus/model/configuration.rb
```diff
@@ -137,6 +144,101 @@ def mapping(path=nil, &blk)
...
+      #
+      # @since 0.3.0
+      # @see Lotus::Model.configure
+      def logging(logger = nil)
+        if logger.nil?
+          @logger
+        else
+          @logger ||= logger
+        end
+      end
```
huydx: what is the difference bw logging and logger?, according to the sample, it's seem to be the same thing

lib/lotus/model/configuration.rb
```diff
@@ -137,6 +144,101 @@ def mapping(path=nil, &blk)
...
+      #
+      # @since 0.3.0
+      # @see Lotus::Model.configure
+      def migration_directory(directory=nil)
+        if directory.nil?
+          @migration_directory ||= _default_migration_directory
+        else
+          @migration_directory = directory
+        end
+      end
```
huydx: migration_directory seem to be both setter and getter, which make me quite confuse.
separate by using different naming seems to be more clearer IMO
```
 Lotus::Model.configure do
   set_migration_directory 'my_custom/migrations'
  end
```

and

AlfonsoUceda: if no directory is passed then `@migration_directory` is setted to default and return this directory, if directory is passed then `@migration_directory` is setted with the passed directory.

lib/lotus/model/configuration.rb
AlfonsoUceda: I agree, I'd renamed to `def logger(value = nil)`

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,287 @@
...
+      # It's raised when adapter is not configured
+      #
+      # @since 0.3.0
+      class MissingAdapterConfiguration < ::StandardError
+        def initialize
+          super('Please configure your adapter. See Lotus::Model.configure for more details')
+        end
+      end
+
+      SUPPORTED_ADAPTERS = [:sql].freeze
```
AlfonsoUceda: document this constant

lib/lotus/model/configuration.rb
vinioliveira: I agree with @huydx, I don't like the idea to be both setter and getter it's a little confused.

lib/lotus/model/configuration.rb
joneslee85: they are pretty much the same. the logging method acts as accessor DSL (though I hate it), I don't mind doing it the Ruby way, that is:

```
logger = Something
```

at all. But I don't want to break the current pattern without running it through with everyone.

lib/lotus/model/configuration.rb
joneslee85: @huydx as much as I hate it, I am hesitated to make changes. I'll have a talk with @jodosha 2moro.

Regarding your suggestion on renaming it to `set_migration_directory`, I don't think it is a good idea for the method is also used as getter.

plukevdh: Can you not follow the standard ruby get/set convention of

`migration_directory` is the getter
`migration_directory=` is the setter

jeremyf: @plukevdh I believe this is a convention that @jodosha has established. Namely, for configuration options, the named configuration option (i.e. `migration_directory`) should be both setter and getter. Consider the following:

```ruby
Lotus::Model.configure do
  migration_directory 'my_custom/migrations'
end
```

If you have a setter, then I believe you will need an explicit receiver, otherwise you are creating a local variable (and not assigning the value to the configuration).

It is a very valid question, and I don't know if @jodosha can point towards a "convention" document. Or if that is something part of the CONTRIBUTING document.

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,290 @@
...
+          raise Lotus::Model::Migrator::UnsupportedAdapter.new(adapter_type)
+        end
+      end
+
+      # Return the adapter configuration
+      #
+      # @since 0.3.0
+      # @api private
+      def _adapter_config
+        configuration.adapter_config
```
vinioliveira: What if we want to use lotus app adapter ? 
Will we should to override the Lotus::Model.configure to use it ?

vinioliveira: Correct if I'm wrong, but Lotus::Application.adapter coexist if Lotus::Model.config.adapter, so if we override Lotus::Model.config.adapter tha previously adapter will be lost. That's the expected behaviour ?

joneslee85: Let me explain, the plan is to copy the configuration from the framework. The configuration is duplicated when calling `.configure`, could be `Lotus::Model.configure` or `MyApp.configure`. Please look into `Lotus::Model::Configuration#duplicate`. Once duplicated, the configuration is then copied to `Lotus::Model::Migrator.configuration` OR `MyApp::Migrator.configuration`. Please see `Lotus::Model::Configuration.for`. I could explain you in details in chat room.

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,290 @@
...
+      # Extract the version from filename
+      #
+      # @example
+      #   _extract_version_from("20150222124516_create_comments.rb")
+      #   # => 20150222124516
+      #
+      # @since 0.3.0
+      # @api private
+      def _extract_version_from(filename)
+        filename.split(::Sequel::Migrator::MIGRATION_SPLITTER, 2).first.to_i
```
AlfonsoUceda: what do you think if this number and above are extracted to constants?

joneslee85: sorry, what number are you implying about?

AlfonsoUceda:
:+1: :heart:

lib/lotus/model/migrator.rb
AlfonsoUceda: number 2, and in `target_version` method, number 1 and 0

lib/lotus/model/configuration.rb
```diff
@@ -182,6 +277,22 @@ def _check_adapter_options!(options)
           raise ArgumentError.new("missing keyword: #{keyword}") if !options.keys.include?(keyword)
         end
       end
+
+      # Default directory that contains migration files
+      #
+      # @since 0.3.0
+      # @api private
+      def _default_migration_directory
+        "#{Dir.pwd}/db/migrations"
```
jodosha: What about `Pathname.pwd.join('db', 'migrations')`? This will make the code OS independent.

lib/lotus/model/adapters/sql/schema_modifier.rb
```diff
@@ -0,0 +1,99 @@
...
+        #   MyMigration.new(adapter).up
+        #
+        #   # to apply down migration
+        #   MyMigration.new(adapter).down
+        #
+        # @since x.x.x
+        # @api private
+        class SchemaModifier
+          def initialize(adapter, logger = nil)
+            @connection = adapter.instance_variable_get(:@connection)
```
jodosha: @joneslee85 Shouldn't this be a public method marked as private API?

lib/lotus/model/adapters/sql/schema_modifier.rb
```diff
@@ -0,0 +1,99 @@
...
+        #
+        #   # to apply up migration
+        #   MyMigration.new(adapter).up
+        #
+        #   # to apply down migration
+        #   MyMigration.new(adapter).down
+        #
+        # @since x.x.x
+        # @api private
+        class SchemaModifier
```
jodosha: @joneslee85 If this class has the pure role of assigning the logger can we get rid of it? Or at least use `SimpleDelegator` for the purpose.

lib/lotus/model/configuration.rb
```diff
@@ -137,6 +146,98 @@ def mapping(path=nil, &blk)
...
+      #   Lotus::Model.configure do
+      #     migration_directory 'my_custom/migrations'
+      #   end
+      #
+      #   Lotus::Model.configuration.migration_directory
+      #     # => 'my_custom/migrations'
+      #
+      # @since 0.3.0
+      # @see Lotus::Model.configure
+      def migration_directory(directory=nil)
```
jodosha: @joneslee85 Shouldn't this be `migrations_directory` instead?

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,544 @@
...
+      # @since x.x.x
+      SCHEMA_TABLE = :schema_migrations
+
+      def initialize(opts={})
+        _copy_framework_configuration
+        _check_if_adapter_is_configured
+        _check_if_adapter_support_migration
+
+        @adapter = _get_adapter
+        @db = adapter.instance_variable_get(:@connection)
```
jodosha: Same as https://github.com/lotus/model/pull/144/files#r29047056

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,544 @@
...
+        @logger = _get_logger
+        @directory = _get_migration_directory
+        @allow_missing_migration_files = opts[:allow_missing_migration_files]
+        @files = _get_migration_files
+        @column = SCHEMA_COLUMN
+        schema, table = @db.send(:schema_and_table, SCHEMA_TABLE)
```
jodosha: This denotes intimate knowledge of `Sequel`. If `@db` isn't our object, we should consider to introduce a `Connection` concept in `Model`. This looks dangerous in the long term.

lib/lotus/model/migrator.rb
```diff
@@ -0,0 +1,544 @@
...
+          yield
+        end
+      end
+
+      # Remove all migration classes.  Done by the migrator to ensure that
+      # the correct migration classes are picked up.
+      #
+      # @since x.x.x
+      # @api private
+      def _remove_migration_classes
```
jodosha: Why this is needed?

joneslee85: Sorry for late reply, it is needed because the way Migration aggregate all existing migration is through the `::Lotus::Model::Migration.descendants` which is appended when a class inherits from `Lotus::Model::Migration`.

lib/lotus/model/migrator.rb
joneslee85: Totally agree, I will work on that

lib/lotus/model/adapters/sql/schema_modifier.rb
joneslee85: @jodosha not really, it's supposed to be the placeholder for our migration helper DSL/methods

deepj:
I may be out but I'd like to suggest to split "schema" and "data" migrations. In Rails days most of developers putting "data" migration into "schema" migration. I often see something like:

```ruby
Model.all.each do |model|
  model.update_attributes(some_attribute: 'something')
end
```

After many years I feel this is very bad approach leading to pain and very long time taking migrations.

I don't have a concrete idea how it would work. It's just a suggestion.

joneslee85:
@deepj yes, thanks for the input. My intent is to only support schema migration, for data migration I'd let user do it with rake task

deepj:
OK, than isn't is better to name the migration class as `Lotus::Model::SchemaMigration` rather than `Lotus::Model::Migration`? I know it's a bit longer. I believe it'll be more obvious for what purposes is meant.

joneslee85:
@deepj I don't think it is necessary. Why? Because we follow the same sequel's class naming and so far it has been used just for schema changes only.

jodosha:
Closed in favor of #196

PR #149, opened 2015-02-06
Removed support for MRI ruby versions lower than 2.2

AlfonsoUceda:
ditto

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,6 @@ sudo: false
 cache: bundler
 script: 'bundle exec rake test:coverage'
 rvm:
-  - 2.0.0
-  - 2.1.0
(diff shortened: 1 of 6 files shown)


jodosha:
@AlfonsoUceda This looks good :+1: We should hold on until the release and keep merging master here.

Can you please remove this line too? https://github.com/lotus/model/blob/0299838bf20c8a1403b50f6cb31245c6e574448b/lib/lotus/entity.rb#L2

joneslee85:
@AlfonsoUceda looks good, can you please also make the params interface of `https://github.com/lotus/model/blob/master/lib/lotus/model/configuration.rb#L97` keyword params? - please remove `_check_adapter_options`

PR #156, opened 2015-02-24
When creating a record, don't send nil values to the database

### LINKED ISSUE #155: Working with default values
Currently `lotus-model` doesn't respect default values in the database:

``` sql
CREATE TABLE deployments (
    id integer NOT NULL,
    project_id integer NOT NULL,
    environment character varying NOT NULL,
    branch character varying NOT NULL,
    deployed_at timestamp with time zone DEFAULT now() NOT NULL
);
```

``` ruby
class Deployment
  include Lotus::Entity
  attributes :id, :project_id, :environment, :branch, :deployed_at
  
  attr_accessor :project
end

deployment = Deployment.new(project_id: 1, environment: 'production', branch: 'master')
DeploymentRepository.persist(deployment)
# Sequel::NotNullConstraintViolation: PG::NotNullViolation: ERROR:  null value in column "deployed_at" violates not-null constraint
# DETAIL:  Failing row contains (4, 1, production, master, null).
```

Seems like the empty `deployed_at` value is being written to the database. I see 2 possible ways to solve this issue: to read the default value into the entity upon it's creation (but I believe it breaks incapsulation and is not very helpful with a default value of `NOW()`), or to save only attributes that were changed.

jodosha:
Closes #155 

/cc @lotus/core-team

### DIFF
--- a/lib/lotus/model/adapters/memory/query.rb
+++ b/lib/lotus/model/adapters/memory/query.rb
@@ -100,7 +100,7 @@ module Lotus
           #        .where(framework: 'lotus')
           def where(condition)
             column, value = _expand_condition(condition)
-            conditions.push([:where, Proc.new{ find_all{|r| r.fetch(column) == value} }])
+            conditions.push([:where, Proc.new{ find_all{|r| r.fetch(column, nil) == value} }])
             self
           end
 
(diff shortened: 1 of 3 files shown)


joneslee85:
LGTM :+1:

7even:
@jodosha thanks, this fixes the problem.

joneslee85:
I close this PR in favour of #161

PR #157, opened 2015-02-25
Fix entity class resolving in Coercer#from_record

7even:
Currently when the name of the entity class clashes with some constant inside `Lotus` namespace the coercer's `from_record` method uses the latter to instantiate a new entity object.

Here's a simple example:

``` ruby
collection :environments do
  entity     ::Environment
  repository ::EnvironmentRepository
  
  attribute :id,   Integer
  attribute :name, String
end

class Environment
  include Lotus::Entity
  attributes :id, :name
end

EnvironmentRepository.first
# => #<Lotus::Environment:0x007fb6e2d53b58
```

The problem lies within instance_eval: the constant inside `@collection.entity` is correct but `instance_eval` evals it inside the `Lotus` namespace so in case of clashes the constant from `Lotus` wins. Prefixing it with double-colon forces `instance_eval` to look for the constant outside.

Feel free to refactor if something is not alright.

### DIFF
--- a/lib/lotus/model/mapping/coercer.rb
+++ b/lib/lotus/model/mapping/coercer.rb
@@ -65,7 +65,7 @@ module Lotus
             end
 
             def from_record(record)
-              #{ @collection.entity }.new(
+              ::#{ @collection.entity }.new(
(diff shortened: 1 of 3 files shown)


joneslee85:
:+1: great thanks

PR #160, opened 2015-03-09
Do not define top-level Boolean, because it is already defined in Utils

janlelis:
I suggest to remove defining the top-level `Boolean` class, since this is already done in [lotus-utils](https://github.com/lotus/utils/commit/139c4d53dbd4914f79187a167c11dd62021b8966), which is a hard dependency of lotus-model.

Having two definitions of `Boolean`
- is surprising
- might lead to surprising behaviour, if they somehow have different implementations


joneslee85:
I like the proposal. What do you think @jodosha?

Besides, do we have to require `lotus/utils` in mapping/collection.rb?

PR #161, opened 2015-03-11
Fix send nil value when creating an entity

AlfonsoUceda:
This is @jodosha 's work, He told me finish the PR:

reference #155 #156

### DIFF
--- a/test/model/mapping/coercer_test.rb
+++ b/test/model/mapping/coercer_test.rb
@@ -9,6 +9,7 @@ describe Lotus::Model::Mapping::Coercer do
     collection.entity(User)
     collection.attribute :id,   Integer
     collection.attribute :name, String
+    collection.attribute :age, String
   end
 
   describe '#to_record' do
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
@lotus/core-team @7even @luishurtado

test/model/mapping/coercer_test.rb
```diff
@@ -16,22 +17,41 @@
...
 
+    describe 'new record' do
+      # Bug:
+      # https://github.com/lotus/model/issues/155
+      it 'ignores unset values' do
+        entity = User.new(name: 'Daenerys Targaryen')
+        coercer.to_record(entity).must_equal(name: 'Daenerys Targaryen')
+      end
+
+      it 'forces nil values' do
```
joneslee85: i am confused with the description, should it be `it ignores nil values on creating?`

AlfonsoUceda: this describe block is when you create new record, we test two cases, new entity without force an attribute to nil and another case forcing a nil attribute.

`to_record` method translate the entity to a format compatible with the database, then nil attributes shouldn't be sent when are creating a entity, but in this case you don't persirst an entity, only we are testing `to_record` method getting the correct values

test/model/mapping/coercer_test.rb
```diff
@@ -16,22 +17,41 @@
...
+        entity = User.new(name: 'Daenerys Targaryen', age: nil)
+        coercer.to_record(entity).must_equal(name: 'Daenerys Targaryen')
```
luishurtado: I think we should not ignore the attributes when explicitly set to nil, even when creating a new record. The test case above should something like:
```ruby
entity = User.new(name: 'Daenerys Targaryen', age: nil)
coercer.to_record(entity).must_equal(name: 'Daenerys Targaryen', age: nil)
```

test/model/mapping/coercer_test.rb
```diff
@@ -16,22 +17,41 @@
       coercer.to_record(entity).must_equal(name: 'Tyrion Lannister')
     end
 
+    describe 'new record' do
+      # Bug:
+      # https://github.com/lotus/model/issues/155
+      it 'ignores unset values' do
+        entity = User.new(name: 'Daenerys Targaryen')
+        coercer.to_record(entity).must_equal(name: 'Daenerys Targaryen')
```
luishurtado: In this test case we should ensure that if a attribute is not explicitly set to nil we wont set it to nil, something like:

```ruby
entity = User.new(name: 'Daenerys Targaryen')
coercer.to_record(entity).wont_equal(name: 'Daenerys Targaryen', age: nil)
```

test/model/mapping/coercer_test.rb
7even: I agree with @luishurtado, that feels less unexpected.

AlfonsoUceda: IMO it doesn't make sense send a nil value to 'database' when you are creating a entity even if you set to nil the attribute.

test/model/mapping/coercer_test.rb
AlfonsoUceda: same as above

test/model/mapping/coercer_test.rb
joneslee85: I am with @AlfonsoUceda on this

test/model/mapping/coercer_test.rb
joneslee85: It's :+1: for the Game of Throne

test/model/mapping/coercer_test.rb
jodosha: @AlfonsoUceda @joneslee85 Sorry but @luishurtado is right.

It may make not sense to send the value for databases with a schema, where the value will remain `nil`/`NULL`.

The issue arises for schema-less databases, when a developer sends the key, but it gets ignored and the structure of the collection isn't altered according their needs. Think of Redis or Mongo in this case.

@AlfonsoUceda Can you please take care of this? Thank you. https://github.com/lotus/model/issues/163

AlfonsoUceda: Ok @jodosha I take care of this ;)

PR #162, opened 2015-03-15
Use modules instead of classes for all repositories

janlelis:
Repositories are not meant to be instantiated, therefore modules should be the encouraged way of creating them!

### DIFF
--- a/EXAMPLE.md
+++ b/EXAMPLE.md
@@ -81,11 +81,11 @@ end
 In order to persist and query the entities above, we define two corresponding repositories:
 
 ```ruby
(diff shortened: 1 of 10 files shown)


jodosha:
@janlelis Thanks for this proposal, but we're still in the process to decide if we want to change the API from class methods to instance methods. Not high priority for now, but this PR will implicitly take a decision as modules can't be instantiated.

Closing this and related PRs.

/cc @lotus/core-team @solnic

janlelis:
Ah, didn't know. Thanks for the explanation.

PR #164, opened 2015-03-16
Fix send nil values to database when explicitly set

### LINKED ISSUE #163: Ensure to send nil values to database when explicitly set
Ref https://github.com/lotus/model/pull/161/files#r26234023
Spec https://gist.github.com/jodosha/a702eff7418f72ba38c7#file-coercer_test-rb-L30

AlfonsoUceda:
fix #163 

@jodosha @luishurtado

what do you think about this approach?

### DIFF
--- a/lib/lotus/entity.rb
+++ b/lib/lotus/entity.rb
@@ -144,7 +144,15 @@ module Lotus
           self.attributes.merge attrs
 
           attrs.each do |attr|
-            attr_accessor(attr) if define_attribute?(attr)
+            if define_attribute?(attr)
+              attr_reader(attr)
+              self.class_eval %Q{
+                def #{attr}=(val)
+                  @initialized_attributes << __callee__
+                  @#{attr} = val
+                end
+              }
+            end
           end
         else
           @attributes ||= Set.new
@@ -181,7 +189,9 @@ module Lotus
     #
     # @see .attributes
     def initialize(attributes = {})
+      @initialized_attributes = Set.new
       attributes.each do |k, v|
+        @initialized_attributes << k
         setter = "#{ k }="
         public_send(setter, v) if respond_to?(setter)
       end
--- a/lib/lotus/model/mapping/coercer.rb
+++ b/lib/lotus/model/mapping/coercer.rb
@@ -61,7 +61,13 @@ module Lotus
                 Hash[#{ @collection.attributes.map{|name,(klass,mapped)| ":#{mapped},Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name})"}.join(',') }]
               else
                 Hash[].tap do |record|
-                  #{ @collection.attributes.reject{|name,_| name == @collection.identity }.map{|name,(klass,mapped)| "value = Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name}); record[:#{mapped}] = value unless value.nil?"}.join('; ') }
+                  #{ @collection.attributes.reject { |name, _| name == @collection.identity }.
+                     map { |name, (klass,mapped)|
+                      %{ value = Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name})
+                         if entity.instance_variable_get(:@initialized_attributes).include?(:#{name})
+                           record[:#{mapped}] = value
+                         end } }.
+                     join('; ') }
                 end
               end
             end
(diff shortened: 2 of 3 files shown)


lib/lotus/entity.rb
```diff
@@ -144,7 +144,15 @@ def attributes(*attrs)
           self.attributes.merge attrs
 
           attrs.each do |attr|
-            attr_accessor(attr) if define_attribute?(attr)
+            if define_attribute?(attr)
+              attr_reader(attr)
+              self.class_eval %Q{
+                def #{attr}=(val)
+                  @initialized_attributes << __callee__
```
joneslee85: @AlfonsoUceda where do you define `@initialized_attributes = Set.new`? It seems to me you need to define it advance just like you do with the constructor.

lib/lotus/model/mapping/coercer.rb
```diff
@@ -61,7 +61,7 @@ def to_record(entity)
                 Hash[#{ @collection.attributes.map{|name,(klass,mapped)| ":#{mapped},Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name})"}.join(',') }]
               else
                 Hash[].tap do |record|
-                  #{ @collection.attributes.reject{|name,_| name == @collection.identity }.map{|name,(klass,mapped)| "value = Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name}); record[:#{mapped}] = value unless value.nil?"}.join('; ') }
+                  #{ @collection.attributes.reject{|name,_| name == @collection.identity }.map{|name,(klass,mapped)| "value = Lotus::Model::Mapping::Coercions.#{klass}(entity.#{name}); record[:#{mapped}] = value if entity.instance_variable_get(:@initialized_attributes).include?(:#{name})"}.join('; ') }
```
joneslee85: I think we should break into lines, this line is getting too hard to follow

lib/lotus/entity.rb
AlfonsoUceda: it's defined in `initialize` method, line 192

lib/lotus/model/mapping/coercer.rb
AlfonsoUceda: I agree, I'll split it

lib/lotus/entity.rb
joneslee85: I get it now

joneslee85:
LGTM :+1:

lib/lotus/entity.rb
```diff
@@ -181,7 +189,9 @@ def inherited(subclass)
     #
     # @see .attributes
     def initialize(attributes = {})
+      @initialized_attributes = Set.new
       attributes.each do |k, v|
+        @initialized_attributes << k
```
jodosha: Should this be guarded by `respond_to?` as well?

```ruby
class Book
  include Lotus::Entity
  attributes :title
end

book = Book.new(foo: 'bar')
book.inspect # => #<Book:0x007fb103cdf1e8 @initialized_attributes=#<Set: {:foo}>>
```

AlfonsoUceda: yes, I agree

jodosha:
@AlfonsoUceda Thank you for taking care of this. I feel uneasy to merge this :sob: Sorry, but it looks hacky :cry: 

We discussed in chat to use `Entity#to_h` to solve the issue. Why this won't work? We should only make sure to extract mapped values from the returned hash.

Pseudocode:

```ruby
def to_record(entity)
  entity.to_h.slice(mapped_attributes)
end
```

Now, `Entity#to_h` returns `Hash`, if we make it return `Utils::Hash` and then implement `#slice`. Not sure if it will work fine, but it's maybe the case to investigate. What do you think?

AlfonsoUceda:
@jodosha yes, it's a little hacky, what do you think some like:
```ruby
def to_record(identity)
  entity.mapped_attributes
end
```
but we need to know the mapped attributes to avoid nils, dont you like an instance storing that?

jodosha:
@AlfonsoUceda I'd avoid to expand Entity's public API, and also because the role of the mapper is to keep entities unaware of persistence. Do we want to avoid `nil` at all? Can we discuss 1:1 on tomorrow? It's quicker. I'll take care of report the results here. :smile:

jodosha:
After a discussion with @AlfonsoUceda we decided to leave things as they are.

We officially support SQL adapter, and the previous fix solved [the problem with null constraints](https://github.com/lotus/model/issues/155). NoSQL adapters aren't officially supported and we don't want to go too far with complexity.

PR #166, opened 2015-03-25
Rename desc() to reverse_order()

joneslee85:


jodosha:
@joneslee85 This seems fair to me, but before to merge we should adjust API docs:

  * `#reverse_order` is `@since x.x.x`
  * `alias_method` is `@since 0.1.0`
  * Add examples to document both `asc` and `desc`

There is an error here: https://github.com/lotus/model/commit/c800a5c6c52a435266776af2d4908c2c1e6b37d1#diff-bcb6af97358a75932ec1322c584f14abR317

Once these issues are solved :+1:

Issue #134, opened 2014-12-26
Add created_at/updated_at timestamp


joneslee85:
What: Having a conventional / implicit timestamp plugin
How:
  * Implement a special attribute helper for collection declaration, eg:
  ```ruby
  collection :users do
    entity     User
    repository UserRepository

    attribute :id,   Integer
    attribute :name, String
    timestamps
  end
  ```

* Enable [Sequel::Plugins::Timestamps](http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/Timestamps.html) If the DB table has column `created_at` and `updated_at`

Pending for discussion

jodosha:
@joneslee85 Doing this via coercer is a bad idea :( It can work when the value is `nil` (`created_at` on a non-persisted entity). But when it comes to update an already existing timestamp (`updated_at`), it won't replace it with a default value.

AlfonsoUceda:
I think the default option should be in the migrations and not in the coercer.

And only care here about include this timestamps

angeloashmore:
@AlfonsoUceda Right, migration-less/schema-less databases (e.g. a JSON store) don't have default values or types. Although there are ways to coerce data to certain data types natively within some of these databases, not every database has that functionality. It would have to take place at the application level.

Is there a reason `#create` only updates the id versus the whole entity reference? My guess is because the database interface only returns that value upon creation?

```ruby
# sql_adapter.rb
def create(collection, entity)
  entity.id = command(
                query(collection)
              ).create(entity)
  entity
end
```

jodosha:
@AlfonsoUceda @angeloashmore This isn't only a matter of schema(less) databases. If a column isn't mapped it's entirely ignored by the framework.

@angeloashmore RE entity: that is my fault. I'm sorry. I already discussed with @joneslee85 about this problem. You have opened a ticket for this, let's discuss over there.

angeloashmore:
It appears the Repository is the best place for this to happen. Maintaining timestamps should occur at `#persist`.

I wrote this quick module and it seems to work. I haven't written any tests for it yet, however.

https://gist.github.com/angeloashmore/05ef7f035f694cfa56d7

## Required setup
1. The database must accept `created_at` and `updated_at`.
2. The entity must respond to `#created_at=` and `#updated_at=`.
3. The repository must set `created_at` and `updated_at` before persisting/creating/updating.
4. The mapper must include `created_at` and `updated_at`.
[...]

PR #169, opened 2015-04-01
Add repository timestamps

dsnipe:
PR based on discussion in #134


lib/lotus/entity.rb
```diff
@@ -138,6 +138,17 @@ module ClassMethods
       #
       #   User.attributes => #<Set: {:id, :name}>
       #   DeletedUser.attributes => #<Set: {:id, :name, :deleted_at}>
+      #
+      # @example Add timestamps
+      #   require 'lotus/model'
+      #
+      #   class User
+      #     include Lotus::Entity
+      #     attributes :timestamps
```
joneslee85: What do you think about having a separate method, for example `add_timestamps` instead? IMHO I think we should allow users to have attribute with naming 'timestamps'.

lib/lotus/model/mapping/collection.rb
```diff
@@ -333,6 +333,29 @@ def attribute(name, klass, options = {})
...
+        #   mapper = Lotus::Model::Mapper.new do
+        #     collection :articles do
+        #       entity Article
+        #
+        #       timestamps
+        #     end
+        #   end
+
+        def timestamps
+          %i(created_at updated_at).each do |name|
```
joneslee85: Nice, I love this %i

lib/lotus/model/mapping/collection.rb
```diff
@@ -333,6 +333,29 @@ def attribute(name, klass, options = {})
           @attributes[name] = [klass, (options.fetch(:as) { name }).to_sym]
         end
 
+        # Define a timestamps for collection.
+        #
+        # `timestamps` helper is translated to:
+        # attribute :created_at, DateTime
+        # attribute :updated_at, DateTime
+        #
+        # @since 0.3.0
```
joneslee85: Please replace with x.x.x

lib/lotus/repository_timestamps.rb
```diff
@@ -0,0 +1,196 @@
+module Lotus
+  module Repository
+    # Timestamps handling using an entity's @created_at and @updated_at.
+    #
+    # @since 0.3.0
+    module Timestamps
+      # Override existing public API into hosting class to support @created_at
+      # and @updated_at using Ruby implementations (database agnostic).
+      #
+      # @since 0.3.0
```
joneslee85: Please replace with x.x.x. For all references in this file too

lib/lotus/entity.rb
joneslee85: Why do we have 2 different ways in defining timestamp? For repo, we do 		+      #     include Lotus::Repository::Timestamps
, why don't we do the same here?

dsnipe:
@joneslee85 thanks for your comments. 
Actually, this feature not finished yet (draft). I decided to first implement a dirty checks, it helps with update_at and then move on.
I have some ideas of realisation now, but need some time to structure it. Today will do and post it for discussion.

lib/lotus/entity.rb
dsnipe: @joneslee85 separate method is good solution, you're right. 
But what do you think about dynamical creation of #created_at and #updated_at? After all, 3 places for declaration timestamps is a little uncomfortable.
Users can declare it in repository and in maps (inside collection). After an entity will be persisted and returned, they'll be have timestamps. 
Also I think about make them read-only for users. What do you think?

dsnipe:
btw, #170 will be very useful here for checking need update #updated_at or not. Thereby, I'll wait till it merged (or declined for some reason) and use it functionality.

joneslee85:
@dsnipe can you please rebase?

dsnipe:
@joneslee85 done
btw, I finish it on this week, I think

joneslee85:
@dsnipe thanks for the update, looking forward to seeing your 2nd PR merge

dsnipe:
@joneslee85, @jodosha 
most work is done, but I have couple thing that should be discussed:
1. In this version, developers should include timestamps modules and helpers in 3 places (repository, in entity and collections). For entities it can be included globally in future in `prepare` configuration block. What do you think about it? It's ok how it is, or maybe somehow create `updated_at` and `created_at` 'magically' in entity after developer included module in Repository and collection? 
2. As dirty tracking is in separate module, I can't manage state of entity, so `updated_at` always will be changed if entity going to persisted, even if it's not changed at all. I can came up with something if we need it (and I think in this case – how to not implement dirty tracking functionality again :) ). Which behavior is correct in this case?

test/repository_timestamps_test.rb
```diff
@@ -0,0 +1,39 @@
...
+  before do
+    adapter, uri, mapper = Lotus::Model::Adapters::SqlAdapter, SQLITE_CONNECTION_STRING, MAPPER
+    ArticleRepository.adapter = adapter.new(mapper, uri)
+  end
+
+  describe 'create timestamp on persisting' do
+    let(:article) { Article.new(title: 'New article with timestamp') }
+    let(:persisted_article) { ArticleRepository.create(article) }
+
+    it '#created_at' do
```
joneslee85: What do you think about using `describe` instead of `it` here? And describe what it does with `it`, for example: `it 'touches created_at attribute'`

test/repository_timestamps_test.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+    it 'change #updated_at on updating' do
+      old_updated_at = persisted_article.updated_at
+      Time.stub :now, Time.now + 10 do
+        article.title = 'Very new article'
+        updated_article = ArticleRepository.persist(article)
+        updated_article.updated_at.to_time.must_be :>, old_updated_at.to_time
+      end
+    end
+    # it 'not change #updated_at'  do
```
joneslee85: unused? remove?

dsnipe: It's related to my question of `updated_at`. I not sure we need it behavior. 
I'll delete it if no need it.

joneslee85:
@dsnipe thanks for this work

> 1. In this version, developers should include timestamps modules and helpers in 3 places (repository, in entity and collections). For entities it can be included globally in future in prepare configuration block. What do you think about it? It's ok how it is, or maybe somehow create updated_at and created_at 'magically' in entity after developer included module in Repository and collection? 

Atm, I have no strong opinion on using global configuration block. @jodosha please weight in your opinion on this

However, I don't think doing magic with entity through repository is a good idea. I want to keep Entity free from Repository knowledge and the other way around.

2. As dirty tracking is in separate module, I can't manage state of entity, so updated_at always will be changed if entity going to persisted, even if it's not changed at all. I can came up with something if we need it (and I think in this case – how to not implement dirty tracking functionality again :) ). Which behavior is correct in this case?

Are you implying about that upon calling `persist` or `update`, though there are no changes, the `updated_at` timestamp is still touched? If so, I am open to suggestion how to NOT implement like dirty tracking (as of Rails for eg).

dsnipe:
> Atm, I have no strong opinion on using global configuration block. @jodosha please weight in your opinion on this

@jodosha write about it here: https://github.com/lotus/model/pull/172#issuecomment-90840706
I like that idea, because it gives opportunity to extend existed functionality of Entities just putting your own module in `prepare` block.  
> However, I don't think doing magic with entity through repository is a good idea. I want to keep Entity free from Repository knowledge and the other way around.

Same for me actually. 
>Are you implying about that upon calling persist or update, though there are no changes, the updated_at timestamp is still touched? If so, I am open to suggestion how to NOT implement like dirty tracking (as of Rails for eg). 

Yep. It tap `updated_at` even if Entity is not changed. I'll think about suggestions and open to ideas to :)

joneslee85:
@dsnipe I think improvement on top this feature can be done as separate PR.

dsnipe:
@joneslee85 agree with you, so leave this PR as it is. 
change naming in tests
did my best, naming is difficult task for me sometimes :)

AlfonsoUceda:
:+1:

lib/lotus/model/mapping/collection.rb
```diff
@@ -334,6 +334,29 @@ def attribute(name, klass, options = {})
...
+        #
+        # @example
+        #   mapper = Lotus::Model::Mapper.new do
+        #     collection :articles do
+        #       entity Article
+        #
+        #       timestamps
+        #     end
+        #   end
+
```
joneslee85: please remove this empty line

lib/lotus/repository/timestamps.rb
```diff
@@ -0,0 +1,196 @@
...
+        # Sets @created_at and @updated_at for the given entity to the current
+        # time in UTC. @created_at is left untouched if already set.
+        #
+        # @param entity [#id] the entity to update
+        #
+        # @return [Object] the entity
+        #
+        # @api private
+        # @since x.x.x
+        def _update_timestamps(entity)
```
joneslee85: FYI #182 has added this update callback in Repository, thus you could remove it

dsnipe:
Finally, I freed some time to continue :)
@joneslee85 after rebase some [tests] (https://github.com/lotus/model/blob/master/test/repository_test.rb#L367-L397) failed.
I suspect that's because in my fixtures I include to Artcile Timestamps module – https://github.com/lotus/model/blob/master/test/fixtures.rb#L8
Not sure what's to do. Easy way is to replace an Article to User class. But maybe better solution is split  fixtures or tests?

jodosha:
@dsnipe Thanks for this PR and I'm terribly sorry for the communication problem that we had here. My apologize :pray: 
I'm closing this, as this feature was already implemented by #173 #176 #182.

/cc @joneslee85

Issue #170, opened 2015-04-01
Dirty checking for entities


dsnipe:
Why not to implement dirty checking for entities?

Something like this:
```ruby
class Article
  include Lotus::Entity
  attributes :title
end

record = Article.new(title: 'My title')
article.changed? # true
article = ArticleRepository.create(record)
article.changed? # false

article.title = 'New title'
article.changed? # true
article.changed_attributes # {:title => 'My title'}
```
I think to put all in 'entity' for now. 
Maybe any thoughts about it?

mvader:
I think it would be really cool to have that feature.

jodosha:
@dsnipe Thanks, I've just merged it! :+1:

jodosha:
Closed by #172

PR #172, opened 2015-04-02
Implementation of dirty check for entity

dsnipe:
#170

### DIFF
--- a/lib/lotus/entity.rb
+++ b/lib/lotus/entity.rb
@@ -138,24 +138,36 @@ module Lotus
       #
       #   User.attributes => #<Set: {:id, :name}>
       #   DeletedUser.attributes => #<Set: {:id, :name, :deleted_at}>
+      #
       def attributes(*attrs)
         if attrs.any?
           attrs = Lotus::Utils::Kernel.Array(attrs)
           self.attributes.merge attrs
 
           attrs.each do |attr|
-            attr_accessor(attr) if define_attribute?(attr)
+            define_attr_accessor(attr) if defined_attribute?(attr)
           end
         else
           @attributes ||= Set.new
         end
       end
 
+      # Create special setter method for attributes.
+      # If value new or changed, set it in hash changed_attributes.
+      #
+      # @params attr [Symbol] an attribute name
+      #
+      # @since x.x.x
+      # @api private
+      def define_attr_accessor(attr)
+        attr_accessor(attr)
+      end
+
       # Check if attr_reader define the given attribute
       #
       # @since 0.2.1
       # @api private
-      def define_attribute?(name)
+      def defined_attribute?(name)
         name == :id ||
           !instance_methods.include?(name)
       end
@@ -237,4 +249,3 @@ module Lotus
 
   end
 end
-
--- a/lib/lotus/entity.rb
+++ b/lib/lotus/entity.rb
@@ -152,8 +152,7 @@ module Lotus
         end
       end
 
-      # Create special setter method for attributes.
-      # If value new or changed, set it in hash changed_attributes.
+      # Define setter/getter methods for attributes.
       #
       # @params attr [Symbol] an attribute name
       #
--- a/lib/lotus/entity/dirty_tracking.rb
+++ b/lib/lotus/entity/dirty_tracking.rb
@@ -23,7 +23,6 @@ module Lotus
       #   article.title = 'Master and Margarita'
       #   article.changed? # => true
       #   article.changed_attributes # => {:title => 'Generation P'}
-      #
       def self.included(base)
         base.class_eval do
           extend ClassMethods
@@ -62,7 +61,6 @@ module Lotus
       # @since x.x.x
       #
       # @see Lotus::Entity#initialize
-      #
       def initialize(attributes = {})
         @changed_attributes = {}
         super
@@ -91,7 +89,6 @@ module Lotus
       #   article.changed_attributes # => {}
       #   article.title = 'Master and Margarita'
       #   article.changed_attributes # => {:title => 'The crime and punishment'}
-      #
       def changed_attributes
         @changed_attributes.dup
       end
@@ -101,7 +98,6 @@ module Lotus
       # @return [TrueClass, FalseClass] the changed indicator
       #
       # @since x.x.x
-      #
       def changed?
         @changed_attributes.size > 0
       end
@@ -121,7 +117,6 @@ module Lotus
       # Clear all information about dirty data
       #
       # @since x.x.x
-      #
       def _clear_changes_information
         @changed_attributes = {}
       end
--- a/lib/lotus/entity/dirty_tracking.rb
+++ b/lib/lotus/entity/dirty_tracking.rb
@@ -44,10 +44,12 @@ module Lotus
         # @api private
         def define_attr_accessor(attr)
           attr_reader(attr)
-          define_method "#{attr}=" do |val|
-            _attribute_changed(attr) if val != __send__(attr)
-            instance_variable_set("@#{attr}", val)
-          end
+          class_eval %{
+            def #{ attr }=(value)
+              _attribute_changed(:#{ attr }) if value != @#{ attr }
+              @#{ attr } = value
+            end
+          }
         end
       end
 
--- a/lib/lotus/entity/dirty_tracking.rb
+++ b/lib/lotus/entity/dirty_tracking.rb
@@ -1,0 +1,129 @@
+module Lotus
+  module Entity
+    # Dirty tracking for entities
+    #
+    # @since x.x.x
+    module DirtyTracking
+      # Override setters for attributes to support dirty tracking
+      #
+      # @since x.x.x
+      #
+      # @example Dirty tracking
+      #   require 'lotus/model'
+      #
+      #   class User
+      #     include Lotus::Entity
+      #     include Lotus::Entity::DirtyTracking
+      #
+      #     attributes :name
+      #   end
+      #
+      #   article = Article.new(title: 'Generation P')
+      #   article.changed? # => false
+      #   article.title = 'Master and Margarita'
+      #   article.changed? # => true
+      #   article.changed_attributes # => {:title => 'Generation P'}
+      #
+      def self.included(base)
+        base.class_eval do
+          extend ClassMethods
+        end
+      end
+
+      module ClassMethods
+        # Override attribute accessors function.
+        # Create setter methods with attribute values checking.
+        # If the new value or a changed, added to @changed_attributes.
+        #
+        # @params attr [Symbol] an attribute name
+        #
+        # @since x.x.x
+        #
+        # @see Lotus::Entity
+        #
+        # @api private
+        def define_attr_accessor(attr)
+          attr_reader(attr)
+          define_method "#{attr}=" do |val|
+            _attribute_changed(attr) if val != __send__(attr)
+            instance_variable_set("@#{attr}", val)
+          end
+        end
+      end
+
+      # Override initialize process.
+      # Init @changed_attributes and Clear dirty data if
+      # entity persisted.
+      #
+      # @param attributes [Hash] a set of attribute names and values
+      #
+      # @since x.x.x
+      #
+      # @see Lotus::Entity#initialize
+      #
+      def initialize(attributes = {})
+        @changed_attributes = {}
+        super
+        _clear_changes_information if self.id
+      end
+
+      # Getter for hash of changed attributes.
+      # Return empty hash, if there is no changes
+      # Getter for hash of changed attributes. Value in it is the previous one.
+      #
+      # @return [::Hash] the changed attributes
+      #
+      # @since x.x.x
+      #
+      # @example
+      #   require 'lotus/model'
+      #
+      #   class User
+      #     include Lotus::Entity
+      #     include Lotus::Entity::DirtyTracking
+      #
+      #     attributes :title
+      #   end
+      #
+      #   article = Article.new(title: 'The crime and punishment')
+      #   article.changed_attributes # => {}
+      #   article.title = 'Master and Margarita'
+      #   article.changed_attributes # => {:title => 'The crime and punishment'}
+      #
+      def changed_attributes
+        @changed_attributes.dup
+      end
+
+      # Return boolean of dirty state for model.
+      #
+      # @return [TrueClass, FalseClass] the changed indicator
+      #
+      # @since x.x.x
+      #
+      def changed?
+        @changed_attributes.size > 0
+      end
+
+      private
+
+      # Set changed attributes in Hash with their old values.
+      #
+      # @params attrs [Symbol] an attribute name
+      #
+      # @since x.x.x
+      # @api private
+      def _attribute_changed(attr)
+        @changed_attributes[attr] = __send__(attr)
+      end
+
+      # Clear all information about dirty data
+      #
+      # @since x.x.x
+      #
+      def _clear_changes_information
+        @changed_attributes = {}
+      end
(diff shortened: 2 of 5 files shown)


lib/lotus/entity.rb
```diff
@@ -144,13 +144,28 @@ def attributes(*attrs)
...
+      # If value new or changed, set it in hash changed_attributes.
+      #
+      # @params attr [Symbol] an attribute name
+      #
+      # @since 0.3.0
+      # @api private
+      def attr_accessor_fabric(attr)
+        attr_reader(attr)
+        define_method "#{attr}=" do |val|
+          _attribute_changed(attr) unless val == __send__(attr)
```
joneslee85: What do you think about `if A != B`?

lib/lotus/entity.rb
```diff
@@ -144,13 +144,28 @@ def attributes(*attrs)
           self.attributes.merge attrs
 
           attrs.each do |attr|
-            attr_accessor(attr) if define_attribute?(attr)
+            attr_accessor_fabric(attr) if define_attribute?(attr)
```
joneslee85: What do you think about renaming it to `define_attr_accessor`?

lib/lotus/entity.rb
```diff
@@ -144,13 +144,28 @@ def attributes(*attrs)
...
           @attributes ||= Set.new
         end
       end
 
+      # Create special setter method for attributes.
+      # If value new or changed, set it in hash changed_attributes.
+      #
+      # @params attr [Symbol] an attribute name
+      #
+      # @since 0.3.0
```
joneslee85: Please change version to `x.x.x`

lib/lotus/entity.rb
```diff
@@ -235,6 +252,46 @@ def update(attributes={})
...
+    #   @changed_attributes # => {:title => 'New title'}
+    #
+    def changed_attributes
+      @changed_attributes ||= Hash.new
+    end
+
+    # Return boolean of dirty state for model.
+    # If model changed, return true, else false.
+    #
+    # @since 0.3.0
```
joneslee85: Please change version to `x.x.x`

lib/lotus/entity.rb
```diff
@@ -235,6 +252,46 @@ def update(attributes={})
       end
     end
 
+    # Getter for hash of changed attributes.
+    # Return empty hash, if there is no changes
+    #
+    # @since 0.3.0
```
joneslee85: Please change version to `x.x.x`

lib/lotus/entity.rb
```diff
@@ -235,6 +252,46 @@ def update(attributes={})
...
+    # @api private
+    def _attribute_changed(attr)
+      @changed_attributes ||= Hash.new
+      @changed_attributes[attr] = __send__(attr)
+    end
+
+    # Clear all dirty data
+    #
+    # @since 0.3.0
```
joneslee85: Please change version to `x.x.x`

lib/lotus/entity.rb
```diff
@@ -235,6 +252,46 @@ def update(attributes={})
...
+      changed_attributes.size > 0
+    end
+
+    private
+
+    # Set changed attributes in Hash with their old values.
+    #
+    # @params attrs [Symbol] an attribute name
+    #
+    # @since 0.3.0
```
joneslee85: Please change version to `x.x.x`

lib/lotus/entity.rb
```diff
@@ -171,6 +186,7 @@ def inherited(subclass)
 
     # Defines a generic, inefficient initializer, in case that the attributes
     # weren't explicitly defined with `.attributes=`.
+    # Clear dirty data if id is set.
```
joneslee85: Do you mean if the entity is persisted?

lib/lotus/entity.rb
```diff
@@ -235,6 +252,46 @@ def update(attributes={})
       end
     end
 
+    # Getter for hash of changed attributes.
+    # Return empty hash, if there is no changes
```
joneslee85: what do you think about following our conventional `@return [::Hash] the changed attributes`?

lib/lotus/entity.rb
```diff
@@ -235,6 +252,46 @@ def update(attributes={})
...
+    #
+    # @example
+    #   @changed_attributes # => {:title => 'New title'}
+    #
+    def changed_attributes
+      @changed_attributes ||= Hash.new
+    end
+
+    # Return boolean of dirty state for model.
+    # If model changed, return true, else false.
```
joneslee85: what do you think about following our conventional `@return [TrueClass, FalseClass] the changed changed indicator

test/entity_test.rb
```diff
@@ -201,4 +201,20 @@ class Camera
...
   end
+
+  describe 'dirty checking' do
+
+    it 'dirty state when created' do
+      book = Book.new(title: 'Crime and Punishment', author: 'F. Dostoevskiy')
+      book.changed?.must_equal true
+    end
+
+    it 'change dirty state' do
```
joneslee85: `s\change\changes`

test/repository_test.rb
```diff
@@ -115,6 +115,12 @@
           UserRepository.create(user1)
           user1.id.must_equal id
         end
+
+        it 'set model as not changed after persisted' do
```
joneslee85: `s\set\sets entity`

joneslee85: `s\not changed\unchanged`

test/repository_test.rb
```diff
@@ -225,6 +231,13 @@ def to_int
           it "returns nil when the given id isn't associated with any entity" do
             UserRepository.find(1_000_000).must_be_nil
           end
+
+          it "not dirty when retrieved from persisted layer" do
```
joneslee85: `s\not\does not`

lib/lotus/entity.rb
joneslee85: on second look, `define_attribute?` is a poor naming, what do you think about renaming it to `defined_attribute?`

joneslee85: Can you please also add additional `@examples` to cover the dirty tracking?

lib/lotus/entity.rb
```diff
@@ -235,6 +252,46 @@ def update(attributes={})
       end
     end
 
+    # Getter for hash of changed attributes.
+    # Return empty hash, if there is no changes
+    #
+    # @since 0.3.0
+    #
+    # @example
+    #   @changed_attributes # => {:title => 'New title'}
```
joneslee85: What do you think about amending the `@example` so that it include entity set/changes of attribute title? I believe it would help user to have a full picture of the workflow.

joneslee85:
@dsnipe overall i am happy with the implementation
@jodosha love to hear your opinion on merging this feature or not

lib/lotus/entity.rb
dsnipe: no problem

lib/lotus/entity.rb
dsnipe: For simple conditions I used unless, but can change it to `!=`

lib/lotus/entity.rb
dsnipe: It only checks for the presence of ID, as keep track of where the model was initialized is not possible yet (from persisted level or by client)
So, if client set ID on init, it clear dirty data too.
Not best solution, but it's only way now to understand that model is persisted. Maybe you have ideas?

dsnipe:
@joneslee85 done with fixes
will use #174 when merged

dsnipe:
@joneslee85 I did rebase from master and used `persisted?` for checking
btw, did you seen my thoughts about `timestamps` realisation? What do you think?

jodosha:
@dsnipe @joneslee85 Thanks for taking care of this. My opinion is that we're recreating AM/AR in Lotus::Model. While I find dirty tracking, timestamps, `#persisted?` useful features I'd not enable them by default.

I'd consider a different approach that consist in different modules like `Lotus::Entity::DirtyTracking`. These modules can be plugged in where needed.

Because this approach would be verbose, we can have a single module to include `Lotus::Entity::Persistence`, which will inject the extra features. We [already use this technique](https://github.com/lotus/helpers/blob/master/lib/lotus/helpers.rb#L20-L27) with `Lotus::Helpers`.

Likewise, this can be noisy too. My last suggestion is to introduce `Lotus::Model::Configuration#prepare_entities`, as a single place where to inject code for **all** the entities. We already use this for Controller and View.

**The bottom line is:** keep entities as simple as possible. This will guarantee speed for the default scenario. When more features are needed, devs can plug-in **one module** in **one place** and they are all set.

@dsnipe Can you please rework this as a module to inject after `Lotus::Entity`? Thank you very much for your effort.

lib/lotus/entity.rb
```diff
@@ -235,6 +267,30 @@ def update(attributes={})
...
+    #     attributes :title
+    #   end
+    #
+    #   article = Article.new(title: 'The crime and punishment')
+    #   article.changed_attributes # => {}
+    #   article.title = 'Master and Margarita'
+    #   article.changed_attributes # => {:title => 'The crime and punishment'}
+    #
+    def changed_attributes
+      @changed_attributes ||= Hash.new
```
jodosha: This isn't thread safe, this ivar should be set when the entity is initialized. Then the getter should be:

```ruby
def changed_attributes
  @changed_attributes.dup
end
```

In this way we prevent accidental modifications of the internal state of the object. Here an example:

```ruby
book.changed_attributes # => {:title => "The crime and punishment"}
book.changed_attributes.delete(:title)
book.changed_attributes # => {}
```

lib/lotus/entity.rb
```diff
@@ -253,5 +309,36 @@ def update(attributes={})
...
+    private
+
+    # Set changed attributes in Hash with their old values.
+    #
+    # @params attrs [Symbol] an attribute name
+    #
+    # @since x.x.x
+    # @api private
+    def _attribute_changed(attr)
+      @changed_attributes ||= Hash.new
```
jodosha: Again, this isn't thread safe.

dsnipe:
@jodosha I had the same feeling about AM/AR, now I starting understand difference and your main idea. Thanks for the good explanation. 
I'll create new module `Lotus::Entity::DirtyTracking`, which can be included by developers in configuration block in future, if they need it. Now I using `persisted?` method. As I understand it will be in separate module, so better not depend on it?

jodosha:
> Now I using persisted? method. As I understand it will be in separate module, so better not depend on it?

Yes. You can safely check for `.id` truthiness. (`if book.id` instead of `if book.persisted?`).

dsnipe:
@jodosha, @joneslee85 
finished with module
can you review it? forgot to make rebase before push, will do it after review with fixes ;)

test/entity_test.rb
```diff
@@ -201,4 +202,26 @@ class Camera
...
+    end
+
+    it 'changes dirty state' do
+      book1 = Book.new
+      book1.changed?.must_equal false
+      book1.title = 'War and Peace'
+      book1.changed?.must_equal true
+      book1.changed_attributes.must_include :title
+    end
+
```
jodosha: Can you please add another test for `#update`? We expect it to already work because it uses internally `public_send("#{attribute}=", value)`. But if we change that implementation, dirty tracking won't work anymore. We need a test to prevent regressions.

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -0,0 +1,129 @@
...
+        # @params attr [Symbol] an attribute name
+        #
+        # @since x.x.x
+        #
+        # @see Lotus::Entity
+        #
+        # @api private
+        def define_attr_accessor(attr)
+          attr_reader(attr)
+          define_method "#{attr}=" do |val|
```
jodosha: Can you please benchmark this implementation and compare it with the following?

```ruby
class_eval %{
  def #{ attr }=(value)
    _attribute_changed(:#{ attr }) if value != @#{ attr }
    @#{ attr } = value
  end
}
```

lib/lotus/entity.rb
```diff
@@ -138,24 +138,35 @@ module ClassMethods
...
       #   DeletedUser.attributes => #<Set: {:id, :name, :deleted_at}>
+      #
       def attributes(*attrs)
         if attrs.any?
           attrs = Lotus::Utils::Kernel.Array(attrs)
           self.attributes.merge attrs
 
           attrs.each do |attr|
-            attr_accessor(attr) if define_attribute?(attr)
+            define_attr_accessor(attr) if defined_attribute?(attr)
```
mvader: Why wrap ```attr_accessor(attr)``` inside ```define_attr_accessor(attr)```? ```define_attr_accessor``` ends up being exactly like ```attr_accessor```. Why not use just ```attr_accessor```?

dsnipe: @mvader because it's easier to override than behaviour; when you inject another module.
you can see it there: https://github.com/dsnipe/model/blob/ac9bcd5ac2612371c214177ae7ab4c5c9598b5ee/lib/lotus/entity/dirty_tracking.rb#L45-L53

dsnipe: I think I need add better explanation in the comments

joneslee85:
@dsnipe can you please rebase?

dsnipe:
@joneslee85 done

AlfonsoUceda:
@dsnipe you have a lot of tests failing, can you take a look?

dsnipe:
@AlfonsoUceda yep, was bad rebase. 
Now all green :)

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -0,0 +1,131 @@
...
+      # Override initialize process.
+      # Init @changed_attributes and Clear dirty data if
+      # entity persisted.
+      #
+      # @param attributes [Hash] a set of attribute names and values
+      #
+      # @since x.x.x
+      #
+      # @see Lotus::Entity#initialize
+      #
```
AlfonsoUceda: remove this line

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -0,0 +1,131 @@
...
+      #     include Lotus::Entity::DirtyTracking
+      #
+      #     attributes :title
+      #   end
+      #
+      #   article = Article.new(title: 'The crime and punishment')
+      #   article.changed_attributes # => {}
+      #   article.title = 'Master and Margarita'
+      #   article.changed_attributes # => {:title => 'The crime and punishment'}
+      #
```
AlfonsoUceda: remove this line

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -0,0 +1,131 @@
...
+      def changed_attributes
+        @changed_attributes.dup
+      end
+
+      # Return boolean of dirty state for model.
+      #
+      # @return [TrueClass, FalseClass] the changed indicator
+      #
+      # @since x.x.x
+      #
```
AlfonsoUceda: remove this line

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -0,0 +1,131 @@
...
+      end
+
+      # Clear all information about dirty data
+      #
+      # @since x.x.x
+      #
+      def _clear_changes_information
+        @changed_attributes = {}
+      end
+
```
AlfonsoUceda: remove this line

AlfonsoUceda:
:+1:

joneslee85:
:+1:

jodosha:
@dsnipe Thank you, just merged :+1:

Issue #167, opened 2015-03-25
Update #created_at when calling Repository#create


joneslee85:
Taker: LeoTeam

* Modify the SQL/Memory adapter so that `Repository#create` updates the `#created_at` if exists
* Ignore if `#created_at` does not exist

PR #173, opened 2015-04-04
Update created at

LeoTeam:
Addressed an issue https://github.com/lotus/model/issues/167

### DIFF
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -673,9 +673,7 @@ module Lotus
 
       def update_timestamps(entity)
         if entity.class.attributes.include?(:created_at)
-          if entity.created_at.nil?
-            entity.created_at = entity.created_at || Time.now.utc
-          end
+          entity.created_at ||= Time.now.utc
         end
       end
 
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -664,9 +664,7 @@ module Lotus
       end
 
       # This is a method to check entity persited or not
-      # 
       # @param entity
-      #
       # @return true
 
       def persisted?(entity)
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -283,7 +283,12 @@ module Lotus
       #
       #   ArticleRepository.create(article) # no-op
       def create(entity)
-        unless entity.id
+        unless 
+          if entity.methods.include?(:created_at)
+            if entity.created_at.nil?
+              entity.created_at = Time.now
+            end
+          end
           @adapter.create(collection, entity)
         end
       end
@@ -379,7 +384,7 @@ module Lotus
       #
       #   ArticleRepository.delete(article) # raises Lotus::Model::NonPersistedEntityError
       def delete(entity)
-        if entity.id
+        if persisted?(entity)
           @adapter.delete(collection, entity)
         else
           raise Lotus::Model::NonPersistedEntityError
@@ -660,6 +665,16 @@ module Lotus
       def exclude(query)
         query.negate!
         query
+      end
+
+      # This is a method to check entity persited or not
+      # 
+      # @param entity
+      #
+      # @return true
+
+      def persisted?(entity)
+        return true if entity.id
       end
     end
   end
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -284,7 +284,7 @@ module Lotus
       #
       #   ArticleRepository.create(article) # no-op
       def create(entity)
-        unless entity.persisted?
+        unless _persisted?(entity)
           _update_timestamps(entity)
           @adapter.create(collection, entity)
         end
@@ -333,7 +333,7 @@ module Lotus
       #
       #   ArticleRepository.update(article) # raises Lotus::Model::NonPersistedEntityError
       def update(entity)
-        if entity.persisted?
+        if _persisted?(entity)
           @adapter.update(collection, entity)
         else
           raise Lotus::Model::NonPersistedEntityError
@@ -381,7 +381,7 @@ module Lotus
       #
       #   ArticleRepository.delete(article) # raises Lotus::Model::NonPersistedEntityError
       def delete(entity)
-        if entity.persisted?
+        if _persisted?(entity)
           @adapter.delete(collection, entity)
         else
           raise Lotus::Model::NonPersistedEntityError
@@ -662,6 +662,14 @@ module Lotus
       def exclude(query)
         query.negate!
         query
+      end
+
+      # This is a method to check entity persited or not
+      # @param entity
+      # @return true
+
+      def _persisted?(entity)
+        !!entity.id
       end
 
       def _update_timestamps(entity)
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -666,8 +666,7 @@ module Lotus
 
       # This is a method to check entity persited or not
       # @param entity
-      # @return true
-
+      # @return a boolean value
       def _persisted?(entity)
         !!entity.id
       end
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -283,7 +283,7 @@ module Lotus
       #
       #   ArticleRepository.create(article) # no-op
       def create(entity)
-        unless persisted?(entity)
+        unless entity.persisted?
           update_timestamps(entity)
           @adapter.create(collection, entity)
         end
@@ -332,7 +332,7 @@ module Lotus
       #
       #   ArticleRepository.update(article) # raises Lotus::Model::NonPersistedEntityError
       def update(entity)
-        if entity.id
+        if entity.persisted?
           @adapter.update(collection, entity)
         else
           raise Lotus::Model::NonPersistedEntityError
@@ -380,7 +380,7 @@ module Lotus
       #
       #   ArticleRepository.delete(article) # raises Lotus::Model::NonPersistedEntityError
       def delete(entity)
-        if persisted?(entity)
+        if entity.persisted?
           @adapter.delete(collection, entity)
         else
           raise Lotus::Model::NonPersistedEntityError
@@ -661,14 +661,6 @@ module Lotus
       def exclude(query)
         query.negate!
         query
-      end
-
-      # This is a method to check entity persited or not
-      # @param entity
-      # @return true
-
-      def persisted?(entity)
-        !!entity.id
       end
 
       def update_timestamps(entity)
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -252,6 +252,7 @@ module Lotus
       #   article = ArticleRepository.find(23)
       #   article.title # => "Launching Lotus::Model"
       def persist(entity)
+        _update_timestamps(entity)
         @adapter.persist(collection, entity)
       end
 
@@ -284,7 +285,7 @@ module Lotus
       #   ArticleRepository.create(article) # no-op
       def create(entity)
         unless entity.persisted?
-          update_timestamps(entity)
+          _update_timestamps(entity)
           @adapter.create(collection, entity)
         end
       end
@@ -663,7 +664,7 @@ module Lotus
         query
       end
 
-      def update_timestamps(entity)
+      def _update_timestamps(entity)
         if entity.class.attributes.include?(:created_at)
           entity.created_at ||= Time.now.utc
         end
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -283,7 +283,7 @@ module Lotus
       #
       #   ArticleRepository.create(article) # no-op
       def create(entity)
-        unless 
+        unless persisted?(entity)
           if entity.methods.include?(:created_at)
             if entity.created_at.nil?
               entity.created_at = Time.now
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -676,7 +676,7 @@ module Lotus
       def update_timestamps(entity)
         if entity.class.attributes.include?(:created_at)
           if entity.created_at.nil?
-            entity.created_at = Time.now
+            entity.created_at = entity.created_at || Time.now.utc
           end
         end
       end
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -284,11 +284,7 @@ module Lotus
       #   ArticleRepository.create(article) # no-op
       def create(entity)
         unless persisted?(entity)
-          if entity.methods.include?(:created_at)
-            if entity.created_at.nil?
-              entity.created_at = Time.now
-            end
-          end
+          update_timestamps(entity)
           @adapter.create(collection, entity)
         end
       end
@@ -674,8 +670,17 @@ module Lotus
       # @return true
 
       def persisted?(entity)
-        return true if entity.id
+        !!entity.id
       end
+
+      def update_timestamps(entity)
+        if entity.class.attributes.include?(:created_at)
+          if entity.created_at.nil?
+            entity.created_at = Time.now
+          end
+        end
+      end
+
     end
   end
 end
--- a/test/repository_test.rb
+++ b/test/repository_test.rb
@@ -57,7 +57,7 @@ describe Lotus::Repository do
         end
 
         describe 'when passed a persisted entity' do
-          let(:user)           { UserRepository.create(User.new(name: 'Don', age: '25')) }
+          let(:user)           { UserRepository.create(User.new(name: 'Don')) }
           let(:persisted_user) { UserRepository.persist(user) }
 
           it 'should return that entity' do
--- a/test/repository_test.rb
+++ b/test/repository_test.rb
@@ -57,6 +57,7 @@ describe Lotus::Repository do
         end
 
         describe 'when passed a persisted entity' do
+
           let(:user)           { UserRepository.create(User.new(name: 'Don', age: '25')) }
           let(:persisted_user) { UserRepository.persist(user) }
 
@@ -90,14 +91,14 @@ describe Lotus::Repository do
 
         describe 'when already persisted' do
           before do
-            @persisted_user = UserRepository.create(User.new(name: 'Don', age: '25'))
-
-            @persisted_user.name = 'Don Dinh'
-            UserRepository.persist(@persisted_user)
+            persisted_user.name = 'Don Dinh'
+            UserRepository.persist(persisted_user)
           end
 
+          let (:persisted_user) { UserRepository.create(User.new(name: 'Don', age: '25')) }
+          
           it 'is updated' do
-            UserRepository.find(@persisted_user.id).must_equal(@persisted_user)
+            UserRepository.find(persisted_user.id).must_equal(persisted_user)
           end
         end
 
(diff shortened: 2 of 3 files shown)


lib/lotus/repository.rb
```diff
@@ -283,7 +283,12 @@ def persist(entity)
       #
       #   ArticleRepository.create(article) # no-op
       def create(entity)
-        unless entity.id
+        unless persisted?(entity)
+          if entity.methods.include?(:created_at)
```
joneslee85: Great work, we can extract this chunk into a private method `update_timestamps`

lib/lotus/repository.rb
```diff
@@ -661,6 +666,16 @@ def exclude(query)
...
       end
+
+      # This is a method to check entity persited or not
+      # 
+      # @param entity
+      #
+      # @return true
+
+      def persisted?(entity)
+        return true if entity.id
```
joneslee85: To ensure the returned value a consistent boolean, you can do this simple trick `!!entity.id`

test/repository_test.rb
```diff
@@ -117,6 +118,34 @@
         end
       end
 
+      describe '.created_at' do
```
joneslee85: By convention, the dot stands for class methods. In our case, we don't test any class method. Thus, please change the description to 'update created_at timestamp' and move this block under `.create` test case.

test/repository_test.rb
```diff
@@ -117,6 +118,34 @@
...
+
+        before do
+
+          @user = UserRepository.create(user1)
+          @timeago = DateTime.new(2015,4,1)
```
joneslee85: `@timeago` is a poor name, what do you think about `created_at`?

test/repository_test.rb
```diff
@@ -117,6 +118,34 @@
...
+        
+        end
+
+        let(:article4) { Article.new(user_id: @user.id, title: 'Love Relationships', comments_count: '4') }
```
joneslee85: Firstly, what do you think about using a simpler fixture entity? Something that has no association, for example User

Secondly, naming with number suffix does not give reader the context of fixtures. What do you think about naming it like  `article_without_created_at` and `article_with_created_at`?

test/repository_test.rb
```diff
@@ -117,6 +118,34 @@
...
+        let(:article5) { Article.new(user_id: @user.id, title: 'Love Relationships', comments_count: '4', created_at: @timeago) }
+
+        it 'persisted created_at which un assign time' do
```
joneslee85: What do you think about renaming it to `assigns and persists created_at attribute`?

test/repository_test.rb
```diff
@@ -117,6 +118,34 @@
...
+          result = ArticleRepository.create(article4)
+          timenow = result.created_at
+          result.instance_variable_get(:@created_at).must_equal timenow
```
joneslee85: It seems to me Article class has not defined `attributes :created_at`. Please double check.

test/repository_test.rb
```diff
@@ -117,6 +118,34 @@
...
+        end
+
+        it 'persisted created_at which assigned time' do
```
joneslee85: Isn't that we already cover this case in the test above?

test/repository_test.rb
joneslee85: Please describe the pre-condition of this test case with an additional `describe 'when entity is already persisted`

test/repository_test.rb
```diff
@@ -117,6 +118,34 @@
...
+          result = ArticleRepository.create(article5)
+          result.instance_variable_get(:@created_at).must_equal @timeago
+        end
+
+        it 'un persist created_at' do
```
joneslee85: What do we assert here? The description is confusing to me.

test/repository_test.rb
mymai91: In file ```test/fixtures.rb``` I added ```attributes :created_at``` in class Article.

mymai91:
@joneslee85 Thanks for your comment. We will update it.

test/repository_test.rb
joneslee85: Then you should be able to call the attribute getter instead of doing the instance variable getter like above

lib/lotus/repository.rb
```diff
@@ -661,6 +662,25 @@ def exclude(query)
...
+      #
+      # @return true
+
+      def persisted?(entity)
+        !!entity.id
+      end
+
+      def update_timestamps(entity)
+        if entity.class.attributes.include?(:created_at)
+          if entity.created_at.nil?
```
joneslee85: on second look, I'd prefer `entity.created_at || Time.now.utc`

test/repository_test.rb
```diff
@@ -18,18 +19,25 @@
       before do
         UserRepository.adapter    = adapter.new(mapper, uri)
         ArticleRepository.adapter = adapter.new(mapper, uri)
-
+
```
joneslee85: remove this blank space

test/repository_test.rb
```diff
@@ -18,18 +19,25 @@
...
-
+        
         UserRepository.collection    = :users
         ArticleRepository.collection = :articles
 
         UserRepository.clear
         ArticleRepository.clear
+
+        unless adapter_name == :file_system
+          PhoneRepository.adapter   = adapter.new(mapper, uri)
```
joneslee85: please align the `=`

test/repository_test.rb
```diff
@@ -18,18 +19,25 @@
...
         ArticleRepository.adapter = adapter.new(mapper, uri)
-
+        
         UserRepository.collection    = :users
         ArticleRepository.collection = :articles
 
         UserRepository.clear
         ArticleRepository.clear
+
+        unless adapter_name == :file_system
```
joneslee85: why do we need this check?

test/repository_test.rb
```diff
@@ -18,18 +19,25 @@
...
+          PhoneRepository.collection   = :phones
+          PhoneRepository.clear
+        end
       end
 
       describe '.collection' do
         it 'returns the collection name' do
           UserRepository.collection.must_equal    :users
           ArticleRepository.collection.must_equal :articles
+          PhoneRepository.collection.must_equal :phones
```
joneslee85: do we need to add this assertion at all? it is not related to our test case

test/repository_test.rb
```diff
@@ -98,6 +106,10 @@
             UserRepository.create(user1),
             UserRepository.create(user2)
           ]
+
+          @created_at = DateTime.new(2015,4,1)
+          @phone_without_created_at = Phone.new(name: 'Iphone 5S', publisher: 'Apple')
```
joneslee85: on second thought, I think I would rename it `unpersisted_phone`?

test/repository_test.rb
```diff
@@ -115,6 +127,23 @@
           UserRepository.create(user1)
           user1.id.must_equal id
         end
+
+        describe 'when entity is already persisted' do
+          it 'assigns and persists created_at attribute' do
+            result = PhoneRepository.create(@phone_without_created_at)
+            time_now = result.created_at
+            result.created_at.must_equal time_now
```
joneslee85: what do you think about combining line 134 and 135?

test/repository_test.rb
mymai91: Oh, Just wanna test this case. I'll remove it.

test/repository_test.rb
```diff
@@ -115,6 +127,23 @@
...
+          end
+
+          it 'assigned and persisted created_at' do
```
joneslee85: how about `it 'does not touch created_at'`?

test/repository_test.rb
```diff
@@ -98,6 +106,10 @@
             UserRepository.create(user1),
             UserRepository.create(user2)
           ]
+
+          @created_at = DateTime.new(2015,4,1)
+          @phone_without_created_at = Phone.new(name: 'Iphone 5S', publisher: 'Apple')
+          @phone_with_created_at = Phone.new(name: 'Galaxy s5', publisher: 'Samsung', created_at: @created_at)
```
joneslee85: this fixture is wrong, it must be setup with `PhoneRepository.persist` so `id` is assigned.

test/repository_test.rb
```diff
@@ -115,6 +127,23 @@
...
+        describe 'when entity is already persisted' do
+          it 'assigns and persists created_at attribute' do
+            result = PhoneRepository.create(@phone_without_created_at)
+            time_now = result.created_at
+            result.created_at.must_equal time_now
+          end
+
+          it 'assigned and persisted created_at' do
+            result = PhoneRepository.create(@phone_with_created_at)
+            result.created_at.must_equal @created_at
```
joneslee85: this assertion is wrong, please fix your fixture to make it a persisted entity by using Repository

test/repository_test.rb
mymai91: Because the `FileSystem::Adapter` require define Phone collection. If not, the test will encounter bug:
```
Lotus::Model::Adapters::FileSystemAdapter::#info#test_0001_returns infos per each collection:
Lotus::Model::Mapping::UnmappedCollectionError: Cannot find collection: phones
```

test/repository_test.rb
```diff
@@ -115,6 +127,23 @@
...
+          end
+        end
+
+        it 'does not exist created_at' do
```
joneslee85: we do not need this test case, please remove it

lib/lotus/repository.rb
```diff
@@ -661,6 +662,25 @@ def exclude(query)
         query.negate!
         query
       end
+
+      # This is a method to check entity persited or not
+      # 
+      # @param entity
+      #
+      # @return true
+
```
joneslee85: Remove empty line

test/repository_test.rb
```diff
@@ -98,6 +105,10 @@
             UserRepository.create(user1),
             UserRepository.create(user2)
           ]
+
+          @created_at        = DateTime.new(2015,4,1)
+          @unpersisted_phone = Phone.new(name: 'Iphone 5S', publisher: 'Apple')
+          @persisted_phone   = Phone.new(name: 'Galaxy s5', publisher: 'Samsung', created_at: @created_at)
```
joneslee85: This is not a persisted entity because it does not have id. Please assign id, just random number

test/repository_test.rb
joneslee85: Can we use User eniry instead of having a new fuxture?

test/repository_test.rb
```diff
@@ -115,6 +126,18 @@
...
           user1.id.must_equal id
         end
+
+        describe 'when entity is already persisted' do
+          it 'assigns and persists created_at attribute' do
+            result = PhoneRepository.create(@unpersisted_phone)
+            result.created_at.must_equal result.created_at
+          end
+
+          it 'does not touch created_at' do
```
joneslee85: Please add another describe line here

test/repository_test.rb
```diff
@@ -115,6 +120,20 @@
           UserRepository.create(user1)
           user1.id.must_equal id
         end
+
+        describe 'when entity is already persisted' do
+          it 'assigns and persists created_at attribute' do
+            result = UserRepository.create(@unpersisted_user)
```
joneslee85: This is wrong, for  unpersisted entity, you don't need to use call repository, just instanstiate the entity

joneslee85: Move this to line 104

test/repository_test.rb
```diff
@@ -115,6 +120,20 @@
...
+            result.created_at.must_equal result.created_at
+          end
+        end
+
+        describe 'when entity is not persisted' do
```
joneslee85: Can we move this to Line 104?

lib/lotus/repository.rb
joneslee85: Sorry what I meant is to remove the check for nil and have this line: `entity.created_at ||= Time.now.utc`

test/repository_test.rb
```diff
@@ -98,6 +98,9 @@
             UserRepository.create(user1),
             UserRepository.create(user2)
           ]
+          
+          @unpersisted_user = User.new(name: 'M', age: '23')
```
joneslee85: Why do we have our fixture declared here instead in a before block at line 123?

test/repository_test.rb
```diff
@@ -115,6 +115,28 @@
...
+            result = UserRepository.create(@unpersisted_user)
+            result.created_at.must_equal result.created_at
+          end
+        end
+
+        describe 'when entity is already persisted' do
+          before do
+            user              = User.new(name: 'D', age: '25')
+            @user             = UserRepository.create(user)
+            @persisted_user   = UserRepository.persist(@user)
```
joneslee85: firstly, why do you use `create` then `persist` here? Should you only use 1 instead?

that is

```
before do
  @persisted_user = UserRepository.persist(User.new(name: 'D', age: '25'))
end

after do
  UserRepository.delete(@persisted_user)
end
```

lib/lotus/repository.rb
```diff
@@ -661,6 +662,21 @@ def exclude(query)
         query.negate!
         query
       end
+
+      # This is a method to check entity persited or not
+      # @param entity
+      # @return true
+
+      def persisted?(entity)
```
joneslee85: you can now remove this method, I've just added `Entity#persisted?` in PR https://github.com/lotus/model/pull/174, so you can rebase with master and start using that method

lib/lotus/repository.rb
```diff
@@ -283,7 +283,8 @@ def persist(entity)
       #
       #   ArticleRepository.create(article) # no-op
       def create(entity)
```
joneslee85: looks good, but have you covered the `Repository#persist` method yet?

test/repository_test.rb
mymai91: *1) Why do you use create then persist here?*

Firstly, I use `.create` to create a User.
After that, I use `.persist` to save User created into DB. 
At the moment, `created_at` will have real value.

*2) Should you only use 1 instead?*
```
before do
  @persisted_user = UserRepository.persist(User.new(name: 'D', age: '25'))
end
```
The way I see that, `User.new(name: 'D', age: '25')` just init a User
After that, you invoke persist method `UserRepository.persist(User.new(name: 'D', age: '25')) `  `created_at` will get value is `nil`. It uncorrect with case `created_at` persisted.

lib/lotus/repository.rb
mymai91: In this issue just update for `create` method so I haven't covered for `persist` yet. I'll do it when I rebase.

test/repository_test.rb
joneslee85: 1) I think you misunderstood the `.persist` and `.create`. Please read the method implementation again.

2) You are testing `.create`, is there any reason why you insist on testing `.persist` now?

Below is how it is supposed to be done:

```ruby
before do
  @persisted_user = UserRepository.persist(User.new(name: 'D', age: '25'))
  @created_at = @persisted_user.created_at
end

after do
  UserRepository.delete(@persisted_user)
end

it 'does not touch created_at' do
  UserRepository.create(@persisted_user)
  @persisted_user.created_at.must_equal @created_at
end
```

lib/lotus/repository.rb
```diff
@@ -283,7 +283,8 @@ def persist(entity)
...
-        unless entity.id
+        unless entity.persisted?
+          update_timestamps(entity)
```
joneslee85: the Lotus conventional naming for private is appending `_` prefix, so please rename this method to `_update_timestamps(entity)`

lib/lotus/repository.rb
```diff
@@ -379,7 +380,7 @@ def update(entity)
       #
       #   ArticleRepository.delete(article) # raises Lotus::Model::NonPersistedEntityError
       def delete(entity)
-        if entity.id
+        if entity.persisted?
```
joneslee85: after talking to @jodosha, it is decide to remove `Entity#persisted?` method because Entity should not be aware of any persistence logic. Please revert it to `entity.id` like before.

mymai91: I got it.

lib/lotus/repository.rb
```diff
@@ -661,6 +663,21 @@ def exclude(query)
         query.negate!
         query
       end
+
+      # This is a method to check entity persited or not
+      # @param entity
+      # @return true
+
```
joneslee85: remove this empty line

lib/lotus/repository.rb
```diff
@@ -661,6 +663,21 @@ def exclude(query)
         query.negate!
         query
       end
+
+      # This is a method to check entity persited or not
+      # @param entity
+      # @return true
```
joneslee85: please see other example on how to write `@return`

test/repository_test.rb
```diff
@@ -90,6 +90,33 @@
             UserRepository.find(@user1.id).must_equal(@user1)
           end
         end
+
+        describe 'when entity is not persisted' do
+          before do
+            @unpersisted_user = User.new(name: 'My', age: '23')
```
joneslee85: there are existing test cases for `.persist` that uses different fixture, can you please consolidate all of them so that all use your newly fixture?

test/repository_test.rb
```diff
@@ -35,61 +35,73 @@
...
+          it 'assigns and persist created_at attribute' do
+            persisted_user = UserRepository.persist(unpersisted_user)
+            persisted_user.created_at.must_equal persisted_user.created_at
           end
         end
 
         describe 'when passed a persisted entity' do
-          let(:user)                   { UserRepository.create(User.new(name: 'S')) }
-          let(:already_persisted_user) { UserRepository.persist(user) }
+
```
joneslee85: remove this blank line

test/repository_test.rb
```diff
@@ -115,6 +127,32 @@
           UserRepository.create(user1)
           user1.id.must_equal id
         end
+
+        describe 'when entity is not persisted' do
+
```
joneslee85: remove empty line

test/repository_test.rb
```diff
@@ -115,6 +127,32 @@
...
+          let(:unpersisted_user) {User.new(name: 'My', age: '23')}
```
joneslee85: add a space after `{`

test/repository_test.rb
```diff
@@ -35,61 +35,73 @@
...
+            UserRepository.persist(persisted_user).must_equal(user)
+          end
+
+          it 'does not touch created_at' do
+            UserRepository.persist(persisted_user)
+            persisted_user.created_at.must_equal persisted_user.created_at
           end
         end
 
         describe 'when non persisted' do
```
joneslee85: from 74 - 103, it is actually a duplicated tests, please remove them

mymai91: I saw it defined before. If it is really unnecessary. I'll remove.

test/repository_test.rb
```diff
@@ -35,59 +35,38 @@
...
+            persisted_user = UserRepository.persist(unpersisted_user)
+            persisted_user.created_at.must_equal persisted_user.created_at
           end
         end
 
-        describe 'when already persisted' do
-          before do
-            @user1 = UserRepository.create(user1)
+        describe 'when passed a persisted entity' do
+          let(:user)           { UserRepository.create(User.new(name: 'Don', age: '25')) }
```
joneslee85: this is wrong...

line 60 and 61 should be consolidated into just 1 line:

```
let(:persisted_user) { UserRepository.create(User.new(name: 'Don')) }
```

mymai91: I will edit `User.new(name: 'Don', age: '25')` into `User.new(name: 'Don')`
But line 60 and 61 should be consolidated
I don't think so. We need to keep it to test for this case ` it 'should return that entity' do `

PR #174, opened 2015-04-05
Add Entity#persisted?

joneslee85:


jodosha:
@joneslee85 Sorry for being late at the party. Can you please consider this design instead? https://github.com/lotus/model/pull/172#issuecomment-90840706

PR #176, opened 2015-04-12
Update updated at

LeoTeam:


test/repository_test.rb
```diff
@@ -68,6 +68,16 @@
             UserRepository.persist(persisted_user)
             persisted_user.created_at.must_equal persisted_user.created_at
           end
+
+          it 'update updated_at' do
```
joneslee85: /s/update/touches

test/repository_test.rb
```diff
@@ -68,6 +68,16 @@
...
+            update_user = UserRepository.find(user.id)
+            update_user.name = 'My'
+
+            UserRepository.persist(update_user)
```
joneslee85: What do you think about reassign the result into the `updated_user` var? That is `updated_user = UserRepository.persist(update_user)`

mymai91: I don't think so. Because `UserRepository.persist(update_user)` it updates an entity. I want to get `updated_user` when I fetch an entity from the collection by its ID.

test/repository_test.rb
```diff
@@ -68,6 +72,17 @@
             UserRepository.persist(persisted_user)
             persisted_user.created_at.must_equal persisted_user.created_at
           end
+
+          it 'touch updated_at' do
```
joneslee85: s/touch/touches

test/repository_test.rb
```diff
@@ -68,6 +72,17 @@
...
+            update_user = UserRepository.find(user.id)
+            update_user.name = 'My'
+
+            UserRepository.persist(update_user)
+
+            updated_user = UserRepository.find(user.id)
```
joneslee85: You can get rid of this extra call by reassigning your variable in line 80

test/repository_test.rb
```diff
@@ -68,6 +72,17 @@
...
+            updated_user.updated_at.must_equal updated_user.updated_at
```
joneslee85: What does this test?

test/repository_test.rb
```diff
@@ -68,6 +72,17 @@
...
+            updated_user.updated_at.wont_equal @updated_at
```
joneslee85: Can we assert that updated_at is newer? Hint: use `assert()`

test/repository_test.rb
```diff
@@ -132,8 +149,10 @@
 
           UserRepository.update(user)
 
-          u = UserRepository.find(@user1.id)
-          u.name.must_equal('Luca')
+          updated_user = UserRepository.find(user.id)
```
joneslee85: Same as above, we can save extra call by reassigning var

test/repository_test.rb
```diff
@@ -132,8 +149,10 @@
...
+          updated_user.name.must_equal('Luca')
+          updated_user.updated_at.must_equal updated_user.updated_at
+          updated_user.updated_at.wont_equal @updated_at
```
joneslee85: Same as above

test/repository_test.rb
mymai91: Got it.

test/repository_test.rb
mymai91: I want to test case updated_at get new value after update entity

test/repository_test.rb
mymai91: I got it. I'll update it.

lib/lotus/repository.rb
```diff
@@ -675,6 +676,13 @@ def _update_timestamps(entity)
         if entity.class.attributes.include?(:created_at)
           entity.created_at ||= Time.now.utc
         end
+        if entity.class.attributes.include?(:updated_at)
+          if entity.updated_at
+            entity.updated_at = Time.now.utc
+          else
+            entity.updated_at = entity.created_at
+          end
+        end
```
AlfonsoUceda: I think you can split this method:
```ruby
def _update_timestamps(entity)
  _update_created_at
  _update_updated_at
```

what do you think?

mymai91: okay.

test/repository_test.rb
```diff
@@ -68,6 +72,17 @@
             UserRepository.persist(persisted_user)
             persisted_user.created_at.must_equal persisted_user.created_at
           end
+
+          it 'touches updated_at' do
+            update_user = UserRepository.find(user.id)
```
joneslee85: what do you think about using `user` directly instead of finding it again?

test/repository_test.rb
```diff
@@ -68,6 +72,17 @@
...
+            update_user.name = 'My'
+
+            updated_user = UserRepository.persist(update_user)
+
+            updated_user = UserRepository.find(user.id)
+            updated_user.updated_at.must_equal updated_user.updated_at
```
joneslee85: what do you think about removing this assertion because it does not really test anything

test/repository_test.rb
mymai91: I agree with you.

test/repository_test.rb
```diff
@@ -50,13 +51,17 @@
             persisted_user.age.must_equal(25)
           end
 
-          it 'assigns and persist created_at attribute' do
+          it 'assigns and persist created_at updated_at attribute' do
             persisted_user = UserRepository.persist(unpersisted_user)
             persisted_user.created_at.must_equal persisted_user.created_at
```
joneslee85: on second look, this assertion does not test any valuable information. What do you think about testing `created_at` not nil instead?

test/repository_test.rb
```diff
@@ -50,13 +51,17 @@
             persisted_user.age.must_equal(25)
           end
 
-          it 'assigns and persist created_at attribute' do
+          it 'assigns and persist created_at updated_at attribute' do
```
joneslee85: good practices is to have 1 assertion per spec example, what do you think about splitting this test into 2?

test/repository_test.rb
```diff
@@ -68,6 +73,12 @@
             UserRepository.persist(persisted_user)
             persisted_user.created_at.must_equal persisted_user.created_at
           end
+
+          it 'touches updated_at' do
+            user.name = 'My'
```
joneslee85: you do not need change any attribute here. Upon calling `persist`, the updated_at will be touched regardlessly

test/repository_test.rb
```diff
@@ -98,9 +109,10 @@
         describe 'when entity is not persisted' do
           let(:unpersisted_user) { User.new(name: 'My', age: '23') }
 
-          it 'assigns and persists created_at attribute' do
+          it 'assigns and persists created_at, updated_at attribute' do
```
joneslee85: what do you think about splitting this into 2?

test/repository_test.rb
```diff
@@ -98,9 +109,10 @@
...
             result = UserRepository.create(unpersisted_user)
             result.created_at.must_equal result.created_at
```
joneslee85: same as above, should be test for non-nil

test/repository_test.rb
mymai91: That's good.

test/repository_test.rb
mymai91: okay

test/repository_test.rb
mymai91: If I don't change any attribute thus according to you does it make sense is "update at  persist method"?

test/repository_test.rb
mymai91: I'll update at here.

test/repository_test.rb
mymai91: I'll update it.

test/repository_test.rb
```diff
@@ -124,16 +142,17 @@
...
         it 'updates entities' do
           user = User.new(name: 'Luca')
           user.id = @user1.id
 
-          UserRepository.update(user)
+          updated_user = UserRepository.update(user)
 
-          u = UserRepository.find(@user1.id)
-          u.name.must_equal('Luca')
+          updated_user.name.must_equal('Luca')
```
joneslee85: split into 2 tests

test/repository_test.rb
```diff
@@ -124,16 +142,22 @@
...
           user = User.new(name: 'Luca')
           user.id = @user1.id
 
-          UserRepository.update(user)
+          updated_user = UserRepository.update(user)
+
+          updated_user.name.must_equal('Luca')
+        end
+
+        it 'updates updated_at attribute' do
```
joneslee85: /s/updates updated_at attribute/touches updated_at

lib/lotus/repository.rb
```diff
@@ -671,12 +672,30 @@ def _persisted?(entity)
...
+      # Add time create an entity
+      def _update_created_at(entity)
         if entity.class.attributes.include?(:created_at)
           entity.created_at ||= Time.now.utc
         end
       end
 
+      # Add time update an entity
+      def _update_updated_at(entity)
+        if entity.class.attributes.include?(:updated_at)
```
jodosha: @mymai91 @joneslee85 This implementation breaks one of the goals of `Lotus::Model`: persist simple objects.

```ruby
class Book
  attr_accessor :id
  end
end

BookRepository.create(Book.new) # => NoMethodError: undefined method `attributes' for Book:Class
```

Can you please add a test case for this scenario and change the implementation to: if `entity.respond_to?(:updated_at=)`? Thank you.

lib/lotus/repository.rb
```diff
@@ -671,12 +672,30 @@ def _persisted?(entity)
...
+          if entity.updated_at
```
jodosha: @mymai91 This can be DRY `entity.updated_at = entity.updated_at || Time.now.utc`

mymai91: `entity.updated_at = entity.updated_at || Time.now.utc`
I think you mean is:
`entity.updated_at = entity.created_at || Time.now.utc`
But The way I see, `entity.updated_at = entity.created_at || Time.now.utc` it won't be appropriate because:
`entity.updated_at = Time.now.utc` if `entity.created_at = nil` but `Time.now.utc` will be assign when we update anything, thereout, `entity.created_at` it can't be nil.
I'll update it to this: `entity.updated_at ? entity.updated_at = Time.now.utc : entity.updated_at = entity.created_at`

PR #181, opened 2015-04-16
Extract push to condition behavior into private method

guilhermeFranco:
Hey guys !

Not sure you are interested in something like this for now, but I noticed that we duplicated the exact same logic (push some data to `conditions`, and exception handling) in three methods after adding expression capability to SQL Adapter. It is good to keep this kind of stuff centralized in one place, for obvious maintainability reasons. 

So, this minor refactor brings this logic into its own private method (it seems the best place for me, since it is totally a helper method) and changes the affected methods to call this new extracted method.

If you guys have any thoughts about this, I'm here to help :yellow_heart: 

Thank you !

### DIFF
--- a/lib/lotus/model/adapters/sql/query.rb
+++ b/lib/lotus/model/adapters/sql/query.rb
@@ -117,9 +117,8 @@ module Lotus
           #   query.where{ age > 10 }
           #
           #   # => SELECT * FROM `users` WHERE (`age` > 31)
-          def where(condition=nil, &blk)
-            condition = (condition or blk or raise ArgumentError.new('You need to specify a condition.'))
-            conditions.push([:where, condition])
+          def where(condition = nil, &blk)
+            _push_to_conditions(:where, condition || blk)
             self
           end
 
@@ -162,9 +161,8 @@ module Lotus
           #   query.where(name: 'John').or{ age > 31 }
           #
           #   # => SELECT * FROM `users` WHERE ((`name` = 'John') OR (`age` < 32))
-          def or(condition=nil, &blk)
-            condition = (condition or blk or raise ArgumentError.new('You need to specify a condition.'))
-            conditions.push([:or, condition])
+          def or(condition = nil, &blk)
+            _push_to_conditions(:or, condition || blk)
             self
           end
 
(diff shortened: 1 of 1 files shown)


lib/lotus/model/adapters/sql/query.rb
```diff
@@ -663,6 +660,11 @@ def apply(query)
             end
           end
 
+          def _push_to_conditions(condition_type, condition)
+            raise ArgumentError, 'You need to specify a condition.' unless condition
```
joneslee85: What do you think about removing this check as Ruby already check for the missing argument

joneslee85:
Personally I am okay with this PR, leave it to @jodosha to decide

lib/lotus/model/adapters/sql/query.rb
AlfonsoUceda: If we remove this raise, then `conditions.push` will add a condition_type with a nil condition, this maybe can be weird and make bugs in the code, because condition can be a condition or a block

guilhermeFranco: I agree with @AlfonsoUceda since `nil || nil == nil`. If we remove the nil check for the second argument we may end with a nil value being propagated to `conditions`

jodosha: @guilhermeFranco Can we change with the following code?

```ruby
raise ArgumentError.new('You need to specify a condition.') if condition.nil?
```

The reasons are:
  1. Better style, passing a message to an instance of an exception is more OOP
  2. In my experience falsey check with Ruby can be source of errors. Explicit `nil` check is more explicit, less bug prone and faster.

Can you please also add API docs for it? Thank you! :+1:

jodosha:
@guilhermeFranco @joneslee85 I left a comment to change a couple of things. After that, :+1: to merge. Thank you!

guilhermeFranco:
@jodosha @joneslee85 Just made the suggested changes. Hope is all good now guys ! Thank you !

lib/lotus/model/adapters/sql/query.rb
```diff
@@ -663,6 +660,23 @@ def apply(query)
...
+          #
+          # @param condition_type [Symbol] the condition type. (eg. `:where`, `:or`)
+          # @param condition [Hash, Proc] the query condition to be stored.
+          #
+          # @return [Array<Array>] the conditions array itself.
+          #
+          # @raise [ArgumentError] if condition is not specified.
+          #
+          # @api private
+          # @since 0.3.1
```
joneslee85: please leave it `x.x.x

lib/lotus/model/adapters/sql/query.rb
```diff
@@ -663,6 +660,23 @@ def apply(query)
...
+          #
```
joneslee85: remove this empty line

joneslee85:
except minor comments above, once done, I'll merge :+1:

guilhermeFranco:
@joneslee85 Done :+1: Thank you !

joneslee85:
@guilhermeFranco thanks for your contribution

PR #182, opened 2015-04-28
Refactor update timestamps

LeoTeam:
* Add test cases for case 'undefined method for Class' at method .persist, .create, .update
* Use respond_to? instead of using attributes.include?
* Refactor code at update timestamps at updated_at

### DIFF
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -672,20 +672,27 @@ module Lotus
         !!entity.id
       end
 
+      # This is a method to check timestamp
+      # @param entity, name
+      # @return a boolean value
+      def _has_timestamp?(entity, name)
+        entity.respond_to?(name) && entity.respond_to?("#{ name }=")
+      end
+
       # Add time create an entity
       def _update_created_at(entity)
-        if entity.class.attributes.include?(:created_at)
+        if _has_timestamp?(entity, :created_at)
           entity.created_at ||= Time.now.utc
         end
       end
 
       # Add time update an entity
       def _update_updated_at(entity)
-        if entity.class.attributes.include?(:updated_at)
-          if entity.updated_at
-            entity.updated_at = Time.now.utc
+        if _has_timestamp?(entity, :updated_at)
+          if entity.respond_to?(:created_at) && entity.created_at
+            entity.updated_at = entity.updated_at ? Time.now.utc : entity.created_at
           else
-            entity.updated_at = entity.created_at
+            entity.updated_at = Time.now.utc
           end
         end
       end
--- a/test/repository_test.rb
+++ b/test/repository_test.rb
@@ -125,7 +125,6 @@ describe Lotus::Repository do
         describe 'when entity is already persisted' do
           before do
             @persisted_user = UserRepository.create(User.new(name: 'My', age: '23'))
-            @created_at     = @persisted_user.created_at
           end
 
           after do
@@ -362,6 +361,39 @@ describe Lotus::Repository do
           @article = ArticleRepository.update(@article)
 
           @article.changed?.must_equal false
+        end
+      end
+
+      describe 'missing timestamps attribute' do
+        describe '.persist' do
+          before do
+            @article = ArticleRepository.persist(Article.new(title: 'Lotus', comments_count: '4'))
+            @article.instance_eval do
+              def created_at
+                @created_at
+              end
+
+              def updated_at
+                @updated_at
+              end
+            end
+          end
+
+          after do
+            ArticleRepository.delete(@article)
+          end
+
+          describe 'when entity does not have created_at accessor' do
+            it 'does not touch created_at' do
+              @article.created_at.must_be_nil
+            end
+          end
+
+          describe 'when entity does not have updated_at accessor' do
+            it 'does not touch updated_at' do
+              @article.updated_at.must_be_nil
+            end
+          end
         end
       end
     end
--- a/test/model/adapters/file_system_adapter_test.rb
+++ b/test/model/adapters/file_system_adapter_test.rb
@@ -43,6 +43,7 @@ describe Lotus::Model::Adapters::FileSystemAdapter do
 
         identity :_id
       end
+      
     end.load!
 
     @adapter = Lotus::Model::Adapters::FileSystemAdapter.new(@mapper, FILE_SYSTEM_CONNECTION_STRING)


lib/lotus/repository.rb
```diff
@@ -674,19 +674,15 @@ def _persisted?(entity)
...
       # Add time update an entity
       def _update_updated_at(entity)
-        if entity.class.attributes.include?(:updated_at)
-          if entity.updated_at
-            entity.updated_at = Time.now.utc
-          else
-            entity.updated_at = entity.created_at
-          end
+        if entity.respond_to?(:updated_at=)
+          entity.updated_at ? entity.updated_at = Time.now.utc : entity.updated_at = entity.created_at
```
joneslee85: please also ensure that entity respond to to `updated_at` and `created_at`

test/repository_test.rb
```diff
@@ -82,6 +82,30 @@
...
+
+            begin
+              CommentRepository.persist(Comment.new)
+            rescue Exception => e
+              @error_class = e.class
+            end
+          end
+
+          it 'persists an entity' do
+            -> { CommentRepository.persist(Comment.new) }.must_raise(@error_class)
```
joneslee85: What do you think about asserting the exception message?

test/repository_test.rb
```diff
@@ -82,6 +82,30 @@
             assert updated_user.updated_at > @updated_at
           end
         end
+
+        describe 'when undefined method for Class' do
```
joneslee85: What do you think about improving this test case message? To me, undefined method is too vague, what method that are undefined? What is Class?

test/repository_test.rb
```diff
@@ -82,6 +82,30 @@
...
+              attr_accessor :id
+              def initialize(attributes = {})
+              end
+            end
+
+            class CommentRepository
+              include Lotus::Repository
+            end
+
+            begin
```
joneslee85: Do we use this block of code at all?

test/repository_test.rb
```diff
@@ -137,6 +161,30 @@
...
```
joneslee85: Same as above

test/repository_test.rb
```diff
@@ -137,6 +161,30 @@
...
+              CommentRepository.create(Comment.new)
+            rescue Exception => e
+              @error_class = e.class
+            end
+          end
+
+          it 'creates an entity' do
+            -> { CommentRepository.create(Comment.new) }.must_raise(@error_class)
```
joneslee85: Same as above

test/repository_test.rb
```diff
@@ -163,6 +211,30 @@
...
+              def initialize(attributes = {})
+              end
+            end
+
+            class CommentRepository
+              include Lotus::Repository
+            end
+
+            begin
+              CommentRepository.update(Comment.new)
```
joneslee85: Same as above

test/repository_test.rb
```diff
@@ -163,6 +211,30 @@
...
+            rescue Exception => e
+              @error_class = e.class
+            end
+          end
+
+          it 'updates an entity' do
+            -> { CommentRepository.update(Comment.new) }.must_raise(@error_class)
```
joneslee85: Same as above

lib/lotus/repository.rb
```diff
@@ -674,19 +674,15 @@ def _persisted?(entity)
 
       # Add time create an entity
       def _update_created_at(entity)
-        if entity.class.attributes.include?(:created_at)
+        if entity.respond_to?(:created_at=)
           entity.created_at ||= Time.now.utc
```
joneslee85: please also check if entity respond to `created_at` or not

test/repository_test.rb
LeoTeam: Can you please tell me why you wanted to raise exception here? 
Because I want to test case error. It will get the error if we don't include attributes when we create a repository.
But now, I got your opinion like that: I just test for case entity have  (updated_at or created_at ) and entity haven't (updated_at or created_at )

lib/lotus/repository.rb
LeoTeam: IMO, attributes in Entity is attr_accessor. So I think I just need to test for a case. I will edit it into `entity.respond_to?(:created_at)` instead of  `entity.respond_to?(:created_at=)`

lib/lotus/repository.rb
mymai91: I think we will have 3 cases:
* Has created_at (Done)
* Has updated_at (Not yet done)
* Have created_at & updated_at (Done & will update condition ensure that entity respond to to `updated_at` and `created_at`)

lib/lotus/repository.rb
```diff
@@ -674,18 +674,18 @@ def _persisted?(entity)
...
       end
 
       # Add time update an entity
       def _update_updated_at(entity)
-        if entity.class.attributes.include?(:updated_at)
-          if entity.updated_at
-            entity.updated_at = Time.now.utc
+        if entity.respond_to?(:updated_at)
+          if entity.respond_to?(:created_at)
+            entity.updated_at ? entity.updated_at = Time.now.utc : entity.updated_at = entity.created_at
```
AlfonsoUceda: shouldn't this method update `update_at` only ?

mymai91: @AlfonsoUceda: Can you make it clearer?

lib/lotus/repository.rb
```diff
@@ -674,18 +674,18 @@ def _persisted?(entity)
 
       # Add time create an entity
       def _update_created_at(entity)
-        if entity.class.attributes.include?(:created_at)
+        if entity.respond_to?(:created_at)
```
joneslee85: need to ensure that `created_at=` is present too

lib/lotus/repository.rb
joneslee85:
```ruby
if entity.respond_to?(:updated_at) && entity.respond_to?(:updated_at=)
  if entity.respond_to?(:created_at) && entity.created_at
    entity.updated_at = !entity.updated_at ? entity.created_at : Time.now.utc
  else
    entity.updated_at = Time.now.utc
  end
end
```

test/repository_test.rb
```diff
@@ -82,6 +82,18 @@
             assert updated_user.updated_at > @updated_at
           end
         end
+
+        describe 'when entity is not respond_to attribute' do
```
joneslee85: s/is not respond_to/does not respond to

test/repository_test.rb
```diff
@@ -82,6 +82,18 @@
...
+          let(:article) { ArticleRepository.persist(Article.new(title: 'Lotus', comments_count: '4')) }
+  
+          it 'not respond_to created_at' do
```
joneslee85: `s/not/does not touch updated_at`

test/repository_test.rb
joneslee85: add another test after the test on the line 52 for the scenario in which `created_at` accessor is not available, we should test for the fact that `created_at` is not going to be touched

joneslee85: add another test after the test on the line 58 for the scenario in which `updated_at` accessor is not available, we should test for the fact that `updated_at` is not going to be touched

test/repository_test.rb
```diff
@@ -82,6 +82,18 @@
             assert updated_user.updated_at > @updated_at
           end
         end
+
```
joneslee85: describe '.persist' do
  end

  describe 'when entity is persisted' do
  end
end

describe 'missing timestamps attribute' do
  describe '.persist' do
    describe 'when entity does not have created_at accessor' do
    describe 'when entity does not have updated_at accessor' do
  end
end

test/repository_test.rb
```diff
@@ -137,6 +148,14 @@
             @persisted_user.created_at.wont_be_nil
           end
         end
+
+        describe 'when entity is not respond_to attribute' do
```
joneslee85: remove this

test/repository_test.rb
```diff
@@ -163,6 +182,45 @@
...
+          end
+
+          after do
+            CommentRepository.clear
+          end
+
+          let(:comment) { CommentRepository.create(Comment.new(comment: 'So great')) }
+
+          it 'not respond_to created_at' do
+            comment.wont_respond_to(:created_at)
```
joneslee85: amend this test, we are not testing if comment responding to created_at or not, but asserting against if updated_at is touched or not

test/repository_test.rb
```diff
@@ -163,6 +182,45 @@
...
+          end
+
+          it 'touches updated_at' do
+            comment.updated_at.wont_be_nil
+          end
+
+        end
+
+        describe 'when entity is not respond_to attribute' do
```
joneslee85: remove this

lib/lotus/repository.rb
```diff
@@ -674,18 +674,18 @@ def _persisted?(entity)
...
       end
 
       # Add time update an entity
       def _update_updated_at(entity)
-        if entity.class.attributes.include?(:updated_at)
-          if entity.updated_at
-            entity.updated_at = Time.now.utc
+        if entity.respond_to?(:updated_at) && entity.respond_to?(:updated_at=)
+          if entity.respond_to?(:created_at) && entity.created_at
+            entity.updated_at ? entity.updated_at = Time.now.utc : entity.updated_at = entity.created_at
```
joneslee85: I think @jodosha would prefer not repeating the `entity.updated_at =`. What do you think about changing it:

```
entity.updated_at = entity.updated_at ? Time.now.utc : entity.created_at
```

test/repository_test.rb
```diff
@@ -163,6 +162,28 @@
         it 'raises an error when not persisted' do
           -> { UserRepository.update(user2) }.must_raise(Lotus::Model::NonPersistedEntityError)
         end
+
+        describe 'when entity contain updated_at without created_at' do
```
joneslee85: s/contain/contains

test/repository_test.rb
```diff
@@ -163,6 +162,28 @@
...
+          before do
+            CommentRepository.adapter    = adapter.new(mapper, uri)
+
```
joneslee85: remove line

lib/lotus/repository.rb
AlfonsoUceda: But why here need the created_at?

do we have difference between created_at and updated_at if we use this?

`entity.updated_at = Time.now.utc`

mymai91: @joneslee85: I got it. It better.

mymai91: @AlfonsoUceda: As you see method update timestamps
```
 def _update_timestamps(entity)
    _update_created_at(entity)
    _update_updated_at(entity)
 end
```

`created_at && updated_at` will get different `time.now.utc` value. Moreover, In the first time, when you create an entity, you don't touch `updated_at` thus `updated_at` can't accept a different value with `created_at`. Hence, the value of `updated_at` must equal `created_at`.

test/repository_test.rb
```diff
@@ -163,6 +162,31 @@
...
+          after do
+            CommentRepository.clear
+          end
+
+          it 'does not touch created_at' do
+            @comment.created_at.must_be_nil
+          end
+
+          it 'touches updated_at' do
+            @comment.updated_at.wont_be_nil
```
joneslee85: we want to ensure that the new value is newer than before, can you please assert that here?

test/repository_test.rb
```diff
@@ -163,6 +162,33 @@
         it 'raises an error when not persisted' do
           -> { UserRepository.update(user2) }.must_raise(Lotus::Model::NonPersistedEntityError)
         end
+
+        describe 'when entity contains updated_at without created_at' do
+          before do
+            CommentRepository.adapter    = adapter.new(mapper, uri)
+            CommentRepository.collection = :comments
+            @comment = CommentRepository.create(Comment.new())
```
joneslee85: /s/new()/new

lib/lotus/repository.rb
```diff
@@ -674,18 +674,18 @@ def _persisted?(entity)
...
           entity.created_at ||= Time.now.utc
         end
       end
 
       # Add time update an entity
       def _update_updated_at(entity)
-        if entity.class.attributes.include?(:updated_at)
-          if entity.updated_at
-            entity.updated_at = Time.now.utc
+        if entity.respond_to?(:updated_at) && entity.respond_to?(:updated_at=)
```
jodosha: @LeoTeam @mymai91 What if we add another method like `_has_timestamp?`, in order to DRY the code? :smile: 

```ruby
def _has_timestamp?(entity, name)
  entity.respond_to?(name) &&
    entity.respond_to?("#{ name }=")
end
```

jodosha:
@LeoTeam @mymai91 This looks good :+1: I left one comment for a refactoring. After that one, can you please squash all the commits in one? Thank you very much! :sparkles: :confetti_ball: 

/cc @joneslee85

lib/lotus/repository.rb
```diff
@@ -672,20 +672,27 @@ def _persisted?(entity)
         !!entity.id
       end
 
+      # This is a method to check timestamp
```
joneslee85: Adding an empty line here

lib/lotus/repository.rb
```diff
@@ -672,20 +672,27 @@ def _persisted?(entity)
...
+      # @param entity, name
+      # @return a boolean value
```
joneslee85: Add `@api private`

lib/lotus/repository.rb
```diff
@@ -672,20 +672,27 @@ def _persisted?(entity)
...
+      def _has_timestamp?(entity, name)
+        entity.respond_to?(name) && entity.respond_to?("#{ name }=")
+      end
+
       # Add time create an entity
       def _update_created_at(entity)
```
joneslee85: Add empty line then following by a `@api private` line

lib/lotus/repository.rb
```diff
@@ -672,20 +672,27 @@ def _persisted?(entity)
...
-        if entity.class.attributes.include?(:created_at)
+        if _has_timestamp?(entity, :created_at)
           entity.created_at ||= Time.now.utc
         end
       end
 
       # Add time update an entity
```
joneslee85: Add empty line then following by a `@api private` line

PR #187, opened 2015-05-17
Fix for Dirty tracking inplace modifications

### LINKED ISSUE #186: Dirty tracking should compare entire object attribute hash
Here's what happens when you bypass lotus attr writers when mutating an enitity:

``` ruby
require 'lotus-model'

class Post
  include Lotus::Entity
  include Lotus::Entity::DirtyTracking
  attributes :title, :tags
end

post = Post.new title: 'foo', tags: %w(red green)

puts post.changed? # false

post.tags << 'blue'
puts post.tags.inspect # ["red", "green", "blue"]

puts post.changed? # false
puts post.changed_attributes.inspect # {}
```

The only way to make it reliable is to compare entire attribute hash, it'll be slower but at least stable.

dsnipe:
It also fixes some wrong tests – `#changed_attributes` should show changed attributes with previous values.

Closes #186

### DIFF
--- a/lib/lotus/entity/dirty_tracking.rb
+++ b/lib/lotus/entity/dirty_tracking.rb
@@ -2,11 +2,11 @@ module Lotus
   module Entity
     # Dirty tracking for entities
     #
-    # @since 0.3.1
+    # @since x.x.x
     module DirtyTracking
-      # Override setters for attributes to support dirty tracking
+      # Support dirty tracking
       #
-      # @since 0.3.1
+      # @since x.x.x
       #
       # @example Dirty tracking
       #   require 'lotus/model'
@@ -25,46 +25,19 @@ module Lotus
       #   article.changed? # => true
       #
       #   article.changed_attributes # => {:title => "Generation P"}
-      def self.included(base)
-        base.class_eval do
-          extend ClassMethods
-        end
-      end
 
-      module ClassMethods
-        # Override attribute accessors function.
-        # Create setter methods with attribute values checking.
-        # If the new value or a changed, added to @changed_attributes.
-        #
-        # @params attr [Symbol] an attribute name
-        #
-        # @since 0.3.1
-        # @api private
-        #
-        # @see Lotus::Entity::ClassMethods#define_attr_accessor
-        def define_attr_accessor(attr)
-          attr_reader(attr)
-
-          class_eval %{
-            def #{ attr }=(value)
-              _attribute_changed(:#{ attr }, @#{ attr }, value)
-              @#{ attr } = value
-            end
-          }
-        end
-      end
+      attr_reader :init_state
 
       # Override initialize process.
       #
       # @param attributes [Hash] a set of attribute names and values
       #
-      # @since 0.3.1
+      # @since x.x.x
       #
       # @see Lotus::Entity#initialize
       def initialize(attributes = {})
-        _clear_changes_information
         super
-        _clear_changes_information
+        @init_state = _current_state
       end
 
       # Getter for hash of changed attributes.
@@ -73,12 +46,12 @@ module Lotus
       #
       # @return [::Hash] the changed attributes
       #
-      # @since 0.3.1
+      # @since x.x.x
       #
       # @example
       #   require 'lotus/model'
       #
-      #   class User
+      #   class Article
       #     include Lotus::Entity
       #     include Lotus::Entity::DirtyTracking
       #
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@dsnipe there is a problem with changed attributes:
```ruby
require 'lotus-model'

class Post
  include Lotus::Entity
  include Lotus::Entity::DirtyTracking
  attributes :title, :tags
end

post = Post.new title: 'foo', tags: %w(red green)

puts post.changed? # false

post.tags << 'blue'
puts post.tags.inspect
post.changed_attributes.inspect # => {:tags=>["red", "green"]}
post.title = "new title"
post.changed_attributes.inspect # => {:title=>"foo"}
```

do you see changed_attributes only save the last changed attribute?

dsnipe:
hmm... I'll check this today
thanks @AlfonsoUceda

dsnipe:
@AlfonsoUceda you may continue reviewing :)

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -2,11 +2,11 @@ module Lotus
   module Entity
     # Dirty tracking for entities
     #
-    # @since 0.3.1
+    # @since x.x.x
```
AlfonsoUceda: I'll keep 0.3.1 version because the public API didn't change

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -2,11 +2,11 @@ module Lotus
...
     module DirtyTracking
-      # Override setters for attributes to support dirty tracking
+      # Support dirty tracking
       #
-      # @since 0.3.1
+      # @since x.x.x
```
AlfonsoUceda: I'll keep 0.3.1 version because the public API didn't change

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -25,46 +25,19 @@ module DirtyTracking
...
-        end
-      end
+      attr_reader :init_state
 
       # Override initialize process.
       #
       # @param attributes [Hash] a set of attribute names and values
       #
-      # @since 0.3.1
+      # @since x.x.x
```
AlfonsoUceda: I'll keep 0.3.1 version because the public API didn't change

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -73,12 +46,12 @@ def initialize(attributes = {})
       #
       # @return [::Hash] the changed attributes
       #
-      # @since 0.3.1
+      # @since x.x.x
```
AlfonsoUceda: I'll keep 0.3.1 version because the public API didn't change

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -91,35 +64,34 @@ def initialize(attributes = {})
...
+        diff = @init_state.to_a - _current_state.to_a
+        Hash[diff]
       end
 
       # Checks if the attributes were changed
       #
       # @return [TrueClass, FalseClass] the result of the check
       #
-      # @since 0.3.1
+      # @since x.x.x
```
AlfonsoUceda: I'll keep 0.3.1 version because the public API didn't change

AlfonsoUceda:
I'll change the position of https://github.com/dsnipe/model/blob/refactoring-dirty-tracking/lib/lotus/entity/dirty_tracking.rb#L7-L27 to https://github.com/dsnipe/model/blob/refactoring-dirty-tracking/lib/lotus/entity/dirty_tracking.rb#L3

lib/lotus/entity/dirty_tracking.rb
```diff
@@ -25,46 +25,19 @@ module DirtyTracking
...
 
-          class_eval %{
-            def #{ attr }=(value)
-              _attribute_changed(:#{ attr }, @#{ attr }, value)
-              @#{ attr } = value
-            end
-          }
-        end
-      end
+      attr_reader :init_state
```
AlfonsoUceda: comment this attr_reader

dsnipe: That how I'm protect initial state (@init_state) of entity from outside modifications. But now I thinking that it can conflict with users methods and better hide it state somehow. 
Or we can put it in public API to show init state (then better to rename it to @init_attributes). 
Also it's possible to implement #revert! that can restore entity to initial state in separate PR (not sure it should be).
What do you think?

AlfonsoUceda: IMO I think you don't protect, an example:
```ruby
post.init_state # => {:id=>nil, :title=>"foo", :tags=>["red", "green"]} 
post.init_state = 123123 #=> NoMethodError: undefined method `init_state=' for #<Post:0x007fefb3a456f0>
post.instance_variable_get(:@init_state) # => {:id=>nil, :title=>"foo", :tags=>["red", "green"]}
post.instance_variable_set(:@init_state, 'fwefwefw')
post.init_state # => "fwefwefw"
```
maybe I'm losing something

dsnipe: You're right. But instance variable always public, so it's possible to protect only getters/setters methods.
Maybe you have any suggestions how to store the initial state in immutable way?

jodosha:
@dsnipe I think this is getting convoluted. The introduction of another ivar to track the state is a smell.
I'd follow @solnic suggestion reported in #186 

> The only way to make it reliable is to compare entire attribute hash, it'll be slower but at least stable.

To clarify, he's referring to `Object#hash` comparison.

solnic:
The only way to make this stable is to compare original "raw state" with the current one. "raw state" means attributes that are going to be persisted. Since entities are decoupled from repositories you're going to have hard time making robust dirty tracking. IMO a repository should be responsible for telling if an object is dirty or not.

dsnipe:
@jodosha I understand what you mean. It's smells for me too. I was looking for a "easy way", but as noted @solnic, it's more complicated than it seems at first glance. 
I'll make research on this when will have some time (after my wedding on this Saturday, I hope at least I'll have it :)

jodosha:
@dsnipe OMG Best wishes!! :wedding: :ring: Don't care about this, I'll figure out. Enjoy your honeymoon :honey_pot:  :full_moon:

dsnipe:
@jodosha thank you! :sunny: 
Working on this project is a pleasure for me, so It'll no problem to take some time. My future wife understands that marries on developer :)
Anyway it'll be great if you can put somewhere your strategic vision about Models (for developing process). Than it'll be easy to understand what functionality important and should be in core, and what is not so and might be as plugin and so on. 
I want to be more involved in that project and that kind of understanding important for me, so I'll be very appreciated if you can find time and share your thoughts.

joneslee85:
@dsnipe Congrats

jodosha:
@dsnipe I merged this thank you. Make it work is more important of the internal design. I've did some refactoring here https://github.com/lotus/model/commit/a230aeaae3a1098136f3d0087074105505a3696f

Issue #190, opened 2015-06-03
Make Repository#execute private


jodosha:
Ditto.

jodosha:
@peterberkenbosch Can you please take care of it? Thank you very much.

mjbellantoni:
@jodosha is this task as simple as moving this method into the private portion of `Repository`?

AlfonsoUceda:
@mjbellantoni yes

mjbellantoni:
I've taken a shot at this in pull #193.

jodosha:
Implemented by #193

PR #193, opened 2015-06-10
Make Respository.execute private (issue #190.)

mjbellantoni:
Makes this method private and removes the associated tests.

### DIFF
--- a/lib/lotus/repository.rb
+++ b/lib/lotus/repository.rb
@@ -551,6 +551,7 @@ module Lotus
         end
       end
 
+      private
       # Executes the given raw statement on the adapter.
       #
       # Please note that it's only supported by some databases,
@@ -595,7 +596,6 @@ module Lotus
         @adapter.execute(raw)
       end
 
-      private
       # Fabricates a query and yields the given block to access the low level
       # APIs exposed by the query itself.
       #
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -48,6 +48,10 @@ class ArticleRepository
   def self.rank_by_user(user)
     rank.by_user(user)
   end
+
+  def self.aggregate
+    execute("select * from articles")
+  end
 end
 
 DB = Sequel.connect(SQLITE_CONNECTION_STRING)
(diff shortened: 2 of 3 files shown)


AlfonsoUceda:
@mjbellantoni ok but you have to test you can't call something like `BookRepository.execute()` and you are able to call execute inside a repository method like:
```ruby
class BookRepository
  include Lotus::Repository
  
  def aggregated
    execute()
  end
```

mjbellantoni:
Can you drop a pointer to an existing test which test something similar?  I'm not that deeply familiar with the framework.  (I'm also not in the habit of writing tests for private methods.)

AlfonsoUceda:
one test can be that you can't call execute like `BookRepository.execute()` another can be use execute inside the example above and assert the response.
```ruby
class BookRepository
  include Lotus::Repository

  def aggregated
    execute("SELECT * FROM books;")
  end
end
```
You can assert that SELECT

jodosha:
@mjbellantoni I think @AlfonsoUceda Is right. Repository has a bunch of **private methods marked as public API**. The reason is that we don't want to leak finders outside of the repository.

The following code isn't possible with Lotus::Model.

```ruby
BookRepository.where(author_id: author.id)
```

This because we want to force devs to define explicit API for their domain model:

```ruby
class BookRepository
  include Lotus::Repository

  def self.by_author(author)
    query do
      where(author_id: author.id)
    end
  end
end
```

Technically, this is a private method, and in general you shouldn't write tests for them. But this is meant to be used by devs (public API), so it deserves to be covered with tests.

Can you please restore and rework them, according to the new semantic? Thank you very much.

mjbellantoni:
I've added the test back in.  There were a couple of places to put these into the test suite.  I left them in the same locations where they were originally.  (The other option was to do something similar to the "querying" test.)

Also, I added a the method `ArticleRepository.aggregate` to that fixture class.  I originally tried adding it to an anonymous subclass of `ArticleRepository` (using `Class.new`) but the class instance variable is not available to subclasses.  (I much prefer that pattern in test suites.)

Looking forward to your feedback!

jodosha:
@mjbellantoni It looks good now. Merged, thank you! :+1:

PR #196, opened 2015-06-11
Database migrations

### LINKED ISSUE #115: Add migration mechanism

jodosha:
## Feature

Introduce **database migrations**.

The following commands are available via `Lotus::Model::Migrator`:

  * `create` - Create database
  * `drop` - Drop database
  * `migrate` - Run migrations
  * `prepare` - Create, load SQL schema (if any), migrate
  * `apply` - Run migrations, dump SQL schema, delete all the migrations (experimental)
  * `version` - Return current database version.

It introduces two extra settings for `Lotus::Model::Configuration`:

  * `migrations` - Migrations directory (defaults to `db/migrations`)
  * `schema` - SQL schema path (defaults to `db/schema.sql`)

---

Please note that `lotusrb` implements **migration generators** and **CLI commands**. This Pull Req defines only **the primitives** that can be used by all Ruby projects that want employ only `lotus-model` as persistence layer. See https://github.com/lotus/lotus/pull/256

---

## Example

```ruby
require 'lotus/model'
require 'lotus/model/migrator'

Lotus::Model.configure do
  # ...
  adapter type: :sql, uri: 'postgres://localhost/bookshelf'
  
  migrations 'db/migrations'
  schema     'db/schema.sql'
end

Lotus::Model::Migrator.create  # => Create database
Lotus::Model::Migrator.migrate # => Run migrations from db/migrations

Lotus::Model::Migrator.migrate(version: "20150610133853")
  # => Migrate to the given version regardless of the direction (up or down)

Lotus::Model::Migrator.drop    # => Drop database
```

## Anatomy Of A Migration

Migrations SHOULD file name must start with a timestamp, followed by an arbitrary snake case name (eg `20150610133853_create_books.rb`). This isn't mandatory, but highly recommended.

Because we wanted to make migrations easy to be handcraft, they use a simple DSL to define a them.

```ruby
Lotus::Model.migration do
  up do
    create_table :books do
      primary_key :id
      column :title, String
    end
  end

  down do
    drop_table :books
  end
end
```

These two blocks (`up/down`) define respectively arbitrary schema changes and their rollback.
Migrations also support an automatic rollback schema changes via `#change`.

```ruby
Lotus::Model.migration do
  change do
    create_table :books do
      primary_key :id
      column :title, String
    end
  end

  # "down" modifications are inferred from `#change` code.
end
```

### Table Definition

A table is defined via `#create_table`. This method accepts an two arguments: the **name** and a **block** that expresses the design.

Safe operation can be performed via `#create_table?`. It only creates the table if not exist.
Force operation can be performed via `#create_table!`. It drops the existing table and creates from scratch.
**These operations shouldn't be used in migrations**.

#### Column Definition

To define a column we use `#column`, followed by the **name**, the **type** and **options**.
The name must be an unique identifier within the table.

The type can be a Ruby type (eg `String`), a symbol that represent a Ruby type (eg `:string`)  , or a string that represents the raw database type (eg `"varchar(255)"`).

##### Type Definition

The following Ruby types are supported:

  * `String` (`varchar(255)`)
  * `Numeric` (`numeric`)
  * `Fixnum` (`integer`)
  * `Integer` (`integer`)
  * `Bignum` (`bigint`)
  * `Float` (`double precision`)
  * `BigDecimal` (`numeric`)
  * `Date` (`date`)
  * `DateTime` (`timestamp`)
  * `Time` (`timestamp`)
  * `TrueClass` (`boolean`)
  * `FalseClass` (`boolean`)
  * `File` (`blob`)

Their translation from Ruby types to database types, may vary from database to database.

##### Options

It supports the following options:

  * `:default` (default value)
  * `:index` (create an index for the column)
  * `:null` (allow NULL values or not)
  * `:primary_key` (make the column primary key for the table)
  * `:unique` (add an uniqueness constraint for the column)

##### Example

```ruby
Lotus::Model.migration do
  change do
    create_table :users do
      primary_key :id
      column :email,     String,  null: false, size: 128, unique: true # VARCHAR(128)
      column :age,       Integer, null: false                          # INT
      column :birthdate, Date                                          # DATE
    end
  end
end
```

#### Primary Key

We can define **primary keys** with the following syntaxes:

```ruby
column :id, Integer, null: false, primary_key: true
# or just use this shortcut
primary_key :id
```

#### Foreign Keys

**Foreign keys** are defined via `#foreign_key`, where we specify the **name** of the column, the **referenced table**, and a set of **options**.
The following example creates an `author_id` column (integer) for `albums` and adds a foreign key.

```ruby
create_table :books do
  # ...
  foreign_key :author_id, :authors, on_delete: :cascade, null: false
end
```

##### Options

It accepts the following options:

  * `:deferrable` (make the constraint check deferrable at the end of a transaction)
  * `:key` (the column in the associated table that this column references. Unnecessary if this column references the primary key of the associated table)
  * `:null` (allow NULL values or not)
  * `:type` (the column type)
  * `:on_delete` (action to take if the referenced record is deleted: `:restrict`, `:cascade`, `:set_null`, or `:set_default`)
  * `:on_update` (action to take if the referenced record is updated: `:restrict`, `:cascade`, `:set_null`, or `:set_default`)

#### Indexes

Indexes are defined via `#index`. It accepts the **name(s)** of the column(s), and a set of **options**.

```ruby
create_table :stores do
  # ...
  column :code, Integer, null: false
  column :lat, Float
  column :lng, Float
  
  index :code, unique: true
  index [:lat, :lng], name: :stores_coords_index
end
```

##### Options

It accepts the following options:

  * `:unique` (uniqueness constraint)
  * `:name` (custom name)
  * `:type` (the type of index, supported by some databases)
  * `:where` (partial index, supported by some databases)

#### Constraints

We can define constraints on columns via `#constraint`. It accepts a **name** and a **block**.

```ruby
create_table :users do
  # ...
  column :age, Integer
  constraint(:adult_constraint) { age > 18 }
end
```

Please note that the block is evaluated in the context of the database engine, **complex Ruby code doesn't work**.
Database functions are mapped to Ruby functions, but this reduces the portability of the migration.

```ruby
create_table :users do
  # ...
  column :password, String
  constraint(:password_length_constraint) { char_length(password) >= 8 }
end
```

#### Checks

Checks are similar to constraints, but they accept an **anoymous block** or a **SQL raw string**.

```ruby
create_table :users do
  # ...
  column :age, Integer
  column :role, String
  
  check { age > 18 }
  check %(role IN("contributor", "manager", "owner"))
end
```

### Table Alterations

The following methods are available for table alterations:

  * `#add_column` (see `#column` for usage)
  * `#drop_column`
  * `#rename_column` (the first argument is the **old name**, while the second is the **new name**)
  * `#add_index` (see `#index` for usage)
  * `#drop_index`
  * `#add_primary_key` (see `#primary_key` for usage)
  * `#add_foreign_key` (see `#foreign_key` for usage)
  * `#add_constraint` (see `#constraint` for usage)
  * `#drop_constraint` (accepts the **name** of the constraint as argument)
  * `#add_unique_constraint`
  * `#set_column_default` (accepts the **name** of the column and the **default value** as comma separated args)
  * `#set_column_type` (accepts the **name** of the column and the **new type** as comma separated args)
  * `#set_column_allow_null` (accepts the **name** of the column)
  * `#set_column_not_null` (accepts the **name** of the column)

Those methods accepts the **name of the target table as first argument**, then the other args.
There is a convenient shortcut for this: `#alter_table`. It accepts the **name of the table** and a **block** that describes the alterations.

The following syntaxes are **equivalent**.

```ruby
Lotus::Model.migration do
  change do
    add_column         :users, :email, String,  null: false, unique: true    
    set_column_default :users, visits_counts, default: 0
  end
end
```

The code above can be DRY'd with:

```ruby
Lotus::Model.migration do
  change do
    alter_table :users do
      # `users` table is implicit within this block, so it can be omitted.
      add_column :email, String,  null: false, unique: true    
      set_column_default visits_counts, default: 0
    end
  end
end
```

### Rename Table

Tables can be renamed via `#rename_table`. It accepts the **old name** and the **new name** as arguments.

```ruby
rename_table :users, :people
```

### Drop Table

Tables can be dropped via `#drop_table`. It accepts the **name** as argument.

```ruby
drop_table :users
```

Safe operation can be performed via `#drop_table?`. It drops the table only if it exists.

## References

Lotus::Model uses [Sequel](http://sequel.jeremyevans.net/) under the hood as database migration engine. If there is any aspect that isn't covered by our documentation or tests, please refer to [Sequel documentation](http://sequel.jeremyevans.net/rdoc/files/doc/schema_modification_rdoc.html).

Thank you @jeremyevans for Sequel :clap: :tophat: :green_heart:

---

Closes #115


celsodantas:
I'm like the way lotus is more modular then Rails, but why the migration is not a separate gem?

jodosha:
@celsodantas For two reasons:

  1. I try to be careful with the amount of dependencies that devs will have to install on their machines, servers and CI. Reducing the number of gems there is less Bundler overhead and faster operations in general.

  2. The long term goal is to extract `lotus-model-sql` from `lotus-model`. The latter will only have memory/file system adapters. All the code related to SQL databases will be handled by that new gem, including the dependency with `sequel`.

celsodantas:
Thanks for the explanation @jodosha, but would be great to have a migration gem independent from Lotus as other ppl could use it without the need to use Lotus.

but again, thanks for the explanation =)

PR #200, opened 2015-06-28
Use createdb to create database

### LINKED ISSUE #199: Invalid connection URI results in PG::ConnectionBad: FATAL: database error
There seems to be a bug when running `lotus db create` and other db-related commands on a new application:

`PG::ConnectionBad: FATAL:  database "Eric" does not exist (Sequel::DatabaseConnectionError)`

I haven't had a chance to dig into it fully, but the problem seems to stem from the following method:

```
 def new_connection
  uri = URI.parse(@connection.uri)
  scheme, userinfo, host, port = uri.select(:scheme, :userinfo, :host, :port)

  uri  = "#{ scheme }://"
  uri += "#{ userinfo }@" unless userinfo.nil?
  uri += host
  uri += ":#{ port }" unless port.nil?

  Sequel.connect(uri)
end
```

It looks like the concatenated `uri` string does not include a database name.  PG is attempting to use the default database `Eric` instead of the configured database name.  I was able to temporarily get around this bug by passing `@connection.uri` to `Sequel.connect`.

/cc @nickcoyne

nickcoyne:
Sequel can’t execute `CREATE` against postgresql. This changes the `create` method to use the `createdb` shell command to create a new database.

Fixes #199

Longer term I think that the that this should be separated out, similar to the [sequel-rails](https://github.com/TalentBox/sequel-rails) gem? Unless of course it has been decided to tie lotus-model into Sequel only. :)

### DIFF
--- a/lib/lotus/model/migrator/postgres_adapter.rb
+++ b/lib/lotus/model/migrator/postgres_adapter.rb
@@ -25,7 +25,7 @@ module Lotus
         # @since 0.4.0
         # @api private
         def create
-          new_connection.run %(CREATE DATABASE "#{ database }"#{ create_options })
+          `createdb #{ database } #{ create_options }`
         end
 
         # @since 0.4.0
@@ -63,7 +63,7 @@ module Lotus
         # @api private
         def create_options
           result  = ""
-          result += %( OWNER "#{ username }") unless username.nil?
+          result += %( --owner=#{ username }) unless username.nil?
           result
         end
 


joneslee85:
Looks good to me. Can you please add test?

lib/lotus/model/migrator/postgres_adapter.rb
```diff
@@ -25,7 +25,7 @@ class PostgresAdapter < Adapter
         # @since 0.4.0
         # @api private
         def create
-          new_connection.run %(CREATE DATABASE "#{ database }"#{ create_options })
```
AlfonsoUceda: why this doesn't run with postgresql? `createdb` is a wrapper for `CREATE DATABASE`

AlfonsoUceda: neverming, I didn't say anything https://github.com/jeremyevans/sequel/issues/877

vyper: `CREATE DATABASE` doesn't work for me also.

jodosha:
@nickcoyne Thanks for this change, merging it. :+1: 

As a side note `CREATE DATABASE` works, but it's less reliable than `createdb`. So this is the way to go.

jodosha:
> Longer term I think that the that this should be separated out, similar to the sequel-rails gem? Unless of course it has been decided to tie lotus-model into Sequel only. :)

@nickcoyne Yes we have [already considered](https://trello.com/c/9QM1Zpna) this. It didn't happened yet because we want to reach first a point where repository and mapping abstractions are mature.

nickcoyne:
@joneslee85 Sorry for not adding tests, I can't actually get the test suite to run locally without issues, before or after this fix.

PR #204, opened 2015-07-03
Removed dump warnings and added name test to database

AlfonsoUceda:
Thanks to @jodosha I added a trick, the test's name to database and I remove two dumps lines in mysql and postgresql because they are unnecessary.

With this PR we avoid these warnings:

`mysqldump: Got error: 1049: Unknown database 'database_name' when selecting the database`
`pg_dump: [archiver (bd)] falló la conexión a la base de datos «database_name»: FATAL:  database "database_name" does not exist`


AlfonsoUceda:
Closing, the failing random test will be fixed in another PR

Issue #191, opened 2015-06-05
Execute raw query postgres


darwinrc:
When trying to execute a raw query through postgresql adapter, the returned value from the method is a Fixnum representing the number of matches instead of the rows. When trying to simulate Sequel behavior as shown in link below, passing a block to the method to cursor the results, that block is not interpreted and a Fixnum is still returned.

The execute method in Sequel can be seen here:

https://github.com/jeremyevans/sequel/blob/98659b904ce13c094c90f59d294a4ba33534030b/lib/sequel/adapters/postgres.rb#L163-L173

jodosha:
@darwinrc Sorry for the late reply on this. This is a good idea. Would you love to implement it yourself? We'll be happy to merge as soon as you'll craft it.

jodosha:
@darwinrc Hello again, I forgot to mention that you can do something like this:

```ruby
result_set = ArticleRepository.execute("SELECT * FROM articles")
result_set.each_hash do |result|
  result # -> { id: 123, title: "Introducing transactions", body: "lorem ipsum"}
end

puts result_set.class # => SQLite3::ResultSet
```

See https://github.com/lotus/model/blob/master/lib/lotus/repository.rb#L589-L592

I'm closing this for now, and open a ticket to promote command/query separation for repositories.

darwinrc:
Thanks a lot @jodosha I'll take that into account.

PR #207, opened 2015-07-10
Implement command/query separation for Repository, by introducing #fetch

jodosha:
### Repository.execute

This PR introduces a breaking change: `Repository.execute` will return `nil`.

This method is still useful if we want to send commands against a database, without expecting any returning value.

```ruby
class ArticleRepository
  include Lotus::Repository

  def self.reset_comment_count
    execute "UPDATE articles SET comments_count = 0"
  end
end
```

### Repository.fetch

In order to retrieve raw records from the database, we have introduced `.fetch`.
**It will return a collection of hashes, representing the RAW records as they are returned from the database, WITHOUT involving coercions and mapper**.

Given the following mapping:

```ruby
collection :articles do
  # ...
  attribute :id, Integer, as: :s_id
  attribute :title, Integer, as: s_title
end
```

It will return

```ruby
class ArticleRepository
  include Lotus::Repository

  def self.find_raw
    fetch "SELECT * FROM articles"
  end
end


ArticleRepository.find_raw
  # => [{:s_id=>"1", :s_title=>"Art 1"}] # They are not coerced or mapped
```

Ref #191 

/cc @lotus/core-team

### DIFF
--- a/lib/lotus/model.rb
+++ b/lib/lotus/model.rb
@@ -28,8 +28,16 @@ module Lotus
     class InvalidMappingError < ::StandardError
     end
 
-    # Error for invalid query
-    # It's raised when a query is malformed
+    # Error for invalid raw command syntax
+    #
+    # @since x.x.x
+    class InvalidCommandError < ::StandardError
+      def initialize(message = "Invalid command")
+        super
+      end
+    end
+
+    # Error for invalid raw query syntax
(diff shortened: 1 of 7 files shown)


darwinrc:
Nice thought taking into account CQRS, @jodosha. Congrats!

Issue #209, opened 2015-07-11
`lotus db drop` fails saying "Cannot find database" (postgres)


ericboehs:
When trying to drop a postgres database, a `Cannot find database` error is thrown.

```
lotus new itshot --database=postgres
cd itshot
lotus db create
lotus db drop
```
Last commands fails with:

```
lotus-model-0.4.1/lib/lotus/model/migrator/postgres_adapter.rb:42:in `rescue in drop':
  Cannot find database: itshot_development (Lotus::Model::MigrationError)
lotus-model-0.4.1/lib/lotus/model/migrator/postgres_adapter.rb:34:in `drop'
lotus-model-0.4.1/lib/lotus/model/migrator.rb:111:in `drop'
lotusrb-0.4.1/lib/lotus/commands/db/drop.rb:9:in `start'
lotusrb-0.4.1/lib/lotus/commands/db.rb:46:in `drop'
thor-0.19.1/lib/thor/command.rb:27:in `run'
thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
thor-0.19.1/lib/thor.rb:359:in `dispatch'
thor-0.19.1/lib/thor/invocation.rb:115:in `invoke'
thor-0.19.1/lib/thor.rb:235:in `block in subcommand'
thor-0.19.1/lib/thor/command.rb:27:in `run'
thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
thor-0.19.1/lib/thor.rb:359:in `dispatch'
thor-0.19.1/lib/thor/base.rb:440:in `start'
lotusrb-0.4.1/bin/lotus:4:in `<top (required)>'
```

I have not tested this on MySQL. Perhaps the same issue is there as well.

ziggurat:
I think this PR will fix this issue https://github.com/lotus/model/pull/213

joneslee85:
@ericboehs to me the problem seems to lie at the create part. There is a now a fix for creation upstream, you can try it out with lotus-model HEAD. 

@ziggurat I tested the drop task for postgresql and it works perfectly.

PR #213, opened 2015-07-14
Replaced DROP DATABASE by dropdb

ziggurat:


ziggurat:
This should fix https://github.com/lotus/model/issues/209

ericboehs:
I don't think this will use the database user specified in the connection url.

ericboehs:
I think we just need to change it to
```ruby
new_connection.run "DROP DATABASE IF EXISTS `#{ database }`"
```

It does seem odd though that it doesn't think the database exists in the case of #209. Is `database` have the correct value? Are the double quotes instead of back ticks throwing it off?

ziggurat:
I can confirm it won't delete the DB even if it exists, I have not checked the double quotes. I made that change because all other commands are being executed using the CLI tools from postgres and not the SQL command. Regarding your comment on using the user, the environment variables are set before running the commands, setting user, host, port and pass. And then those environment variables are being picked up by the createdb command as stated in the documentation:

> PGHOST
PGPORT
PGUSER
Default connection parameters
This utility, like most other PostgreSQL utilities, also uses the environment variables supported by libpq

Nevertheless, I will chek the double quotes.
Thanks!

ziggurat:
Had no luck changing the quotes.

AlfonsoUceda:
@ziggurat did you try with @ericboehs suggestion? http://www.postgresql.org/docs/9.1/static/sql-dropdatabase.html

ziggurat:
@AlfonsoUceda yes, I tried that solution without luck. I have some time today, I will try again and let you know.

ziggurat:
@ericboehs @AlfonsoUceda any ideas? As stated in the posgres documentation, we can't run the DROP command connected to the DB we want to drop. So I think alternatives are using a different connection to run the DROP command or using the dropdb postgres script.

joneslee85:
@ziggurat can you please add test? I am not convinced till we have test

ziggurat:
I am not sure what test to add, tests are creating and droping dbs succesfully using the migrators. There is 100% coverage for the modified adapter. Sorry for being a pain, it's my first time collaborating in an open source project.

joneslee85:
@ziggurat I've just done a hand-test and I could be able to drop DB without the needs of this patch.

joneslee85:
@ziggurat I close this ticket now because your edge case will never happen, because we instantiate new connection for every drop query and this connection does not connect to the database

joneslee85:
I amended your PR to keep the old behaviour

https://github.com/lotus/model/pull/223/files

PR #214, opened 2015-07-20
Fix documentation for create/persist.

pascalbetz:
create/persist do not assign the id to the entity passed as parameter.
Instead they return a copy of the entity with id assigned.


joneslee85:
:+1:

AlfonsoUceda:
:+1:

PR #216, opened 2015-07-27
Implemented mapper custom coercer

### LINKED ISSUE #183: Array Coercion makes a 'record' not a text[] in pg
Using postgresql and lotus model from the master branch. I would like to store an array of Strings in an entity like this:

```ruby
attribute :images, type: Array
```
Then migrate the database with this:
```ruby
add_column :products, :images, "text[]"
```

I verified that the database was migrated correctly. 

When I try to run the specs storing an array of images like this:

```ruby
images: [
          "http://dummyimage.com/600x400/000/fff&text=1",
          "http://dummyimage.com/600x400/000/fff&text=2",
          "http://dummyimage.com/600x400/000/fff&text=3"
        ]
```

I get this error from Sequel:

```
Sequel::DatabaseError:
       PG::DatatypeMismatch: ERROR:  column "images" is of type text[] but expression is of type record
       LINE 1: ....814681+0000', '2015-05-04 10:08:00.814681+0000', ('http://d...
                                                                    ^
       HINT:  You will need to rewrite or cast the expression.
```

I would love follow that HINT, and I will try to. Not 100% sure if this is something in Lotus currently of just me using the attributes in a wrong way.

jodosha:
Allow developers to specify a custom coercer.

### Example

Given the following Postgres table:

```sql
CREATE TABLE articles (
  id     integer NOT NULL,
  title  varchar(128),
  tags   text[],
);
```

And the following mapping:

```ruby
require 'lotus/model/coercer'
require 'sequel/extensions/pg_array'

class PGArray < Lotus::Model::Coercer
  def self.dump(value)
    ::Sequel.pg_array(value) rescue nil
  end

  def self.load(value)
    ::Kernel.Array(value) unless value.nil?
  end
end

Lotus::Model.configure do
  mapping do
    collection :articles do
      attribute :id,    Integer
      attribute :title, String
      attribute :tags,  PGArray
    end
  end
end
```

#### Persist A Record

When the entity is persisted as a record into the database:

  * `title` is coerced via `Lotus::Model::Mapping::Coercers::String.dump` which returns a `String`
  * `tags` is coerced via `PGArray.dump` which returns a string that is compatible with Postgres array syntax.

#### Fetch A Record

When the entity is fetched from the database:

  * `title` is coerced via `Lotus::Model::Mapping::Coercers::String.load` which returns a `String`
  * `tags` is coerced via `PGArray.load` which returns an `Array`

Closes #183 

/cc @lotus/core-team


AlfonsoUceda:
:+1:

jodosha:
This allows me to use PG arrays right now, in an application that I'm building.

lib/lotus/model/mapping/attribute.rb
```diff
@@ -0,0 +1,105 @@
...
+        #
+        # @api private
+        # @since x.x.x
+        #
+        # @see Lotus::Model::Mapping::Coercions
+        # @see Lotus::Model::Mapping::Collection#attribute
+        def initialize(name, klass, options)
+          @name    = name.to_sym
+          @klass   = klass
+          @options = options
```
joneslee85: what do you think about adding attr_reader for this var?

joneslee85:
:+1:

lib/lotus/model/mapping/attribute.rb
AlfonsoUceda: @joneslee85 why do we need that option?

lib/lotus/model/mapping/coercer.rb
```diff
@@ -58,17 +58,17 @@ def deserialize_#{ mapped }(value)
           instance_eval <<-EVAL, __FILE__, __LINE__
```
brennovich: Non related question: is there a way to use `define_method` here http://tenderlovemaking.com/2013/03/03/dynamic_method_definitions.html? Also, I put in my TODO list to simplify this ;)

lib/lotus/model/mapping/attribute.rb
jodosha: @joneslee85 yup, what's the purpose of that? :smile:

AlfonsoUceda:
to me :ok_hand:

PR #222, opened 2015-08-04
Added group query to sql

AlfonsoUceda:
ditto

I added group query, `having` will be added in other PR.

Now you can use `GROUP BY` query with the `group` method.

```ruby
class User
  include Lotus::Entity
  attributes :name, :age
end

class UserRepository
  def grouped_by_age
    query do
      group(:age)
    end
  end
end
```

### DIFF
--- a/lib/lotus/model/adapters/memory/query.rb
+++ b/lib/lotus/model/adapters/memory/query.rb
@@ -541,6 +541,19 @@ module Lotus
             raise NotImplementedError
           end
 
+          # This method is defined in order to make the interface of
+          # `Memory::Query` identical to `Sql::Query`, but this feature is NOT
+          # implemented
+          #
+          # @raise [NotImplementedError]
+          #
+          # @since x.x.x
+          #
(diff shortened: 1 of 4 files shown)


joneslee85:
:+1:

AlfonsoUceda:
@darwinrc did this pr be useful for you?

jodosha:
@AlfonsoUceda Lovely :+1:

lib/lotus/model/adapters/memory/query.rb
```diff
@@ -541,6 +541,19 @@ def negate!
...
+          # `Memory::Query` identical to `Sql::Query`, but this feature is NOT
+          # implemented
+          #
+          # @raise [NotImplementedError]
+          #
+          # @since x.x.x
+          #
+          # @see Lotus::Model::Adapters::Sql::Query#group!
+          def group
+            raise NotImplementedError
```
AlfonsoUceda: @jodosha what do you think to add group by for memory adapter? ruby brings us that method.

PR #223, opened 2015-08-04
Use dropdb to drop DB with PostgreSQL

joneslee85:


lib/lotus/model/migrator/postgres_adapter.rb
```diff
@@ -32,15 +32,25 @@ def create
...
-          end
+          set_environment_variables
+
+          require 'open3'
+
+          Open3.popen3('dropdb', database) do |stdin, stdout, stderr, wait_thr|
+            exit_status = wait_thr.value
 
-          raise MigrationError.new(message)
+            if !exit_status.success?
```
AlfonsoUceda: replace `if !` with `unless`

PR #228, opened 2015-08-16
Sql joins

AlfonsoUceda:
## SQL JOINS
Support for `INNER JOIN` and `LEFT JOIN` sql operations

#### INNER JOIN
```ruby
# Assert we have articles and users tables
require 'lotus/model'

class UserRepository
  include Lotus::Repository

  def self.users_with_articles
    query do
      join(:articles)
    end
  end
end
```
#### LEFT JOIN
```ruby
# Assert we have articles and users tables
require 'lotus/model'

class UserRepository
  include Lotus::Repository

  def self.users_with_and_without_articles
    query do
      left_join(:articles)
    end
  end
end
```

### DIFF
--- a/test/model/adapters/sql_joins_test.rb
+++ b/test/model/adapters/sql_joins_test.rb
@@ -1,0 +1,309 @@
+require 'test_helper'
+require 'lotus/model/migrator'
+
+describe 'SQL joins test' do
+  before do
+    class TestUserJoin
+      include Lotus::Entity
+
+      attributes :country_id, :name, :age
+    end
+
+    class TestUserJoinRepository
+      include Lotus::Repository
+    end
+
+    class TestDeviceJoin
+      include Lotus::Entity
+
+      attributes :u_id
+    end
+
+    class TestDeviceJoinRepository
+      include Lotus::Repository
+    end
+
+    class TestOrder
+      include Lotus::Entity
+
+      attributes :user_id, :total
+    end
+
+    class TestOrderRepository
+      include Lotus::Repository
+    end
+
+    class TestAge
+      include Lotus::Entity
+
+      attributes :value, :label
+    end
+
+    class TestAgeRepository
+      include Lotus::Repository
+    end
+
+    class TestCountry
+      include Lotus::Entity
+
+      attributes :code, :country_id
+    end
+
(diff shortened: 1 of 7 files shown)


test/model/adapters/sql_joins_test.rb
```diff
@@ -0,0 +1,311 @@
...
+              @adapter.create(:orders, @order1)
+              @adapter.create(:orders, @order2)
+              @adapter.create(:orders, @order3)
+
+              TestUserRepository.adapter = @adapter
+              TestOrderRepository.adapter = @adapter
+            end
+
+            it 'returns records' do
+              created_user = TestUserRepository.first
```
jodosha: To be sure we're referring to the same record, we should use `@created_user` from the `before` block, instead of finding it again. :wink:

test/model/adapters/sql_joins_test.rb
```diff
@@ -0,0 +1,311 @@
...
+
+              user = TestUser.new(country_id: country2.id)
+              created_user = @adapter.create(:users, user)
+
+              TestUserRepository.adapter = @adapter
+              TestCountryRepository.adapter = @adapter
+            end
+
+            it 'returns records' do
+              created_user = TestUserRepository.first
```
jodosha: Same as: https://github.com/lotus/model/pull/228/files#r39265171

test/model/adapters/sql_joins_test.rb
```diff
@@ -0,0 +1,311 @@
...
+
+              @adapter.create(:devices, @device1)
+              @adapter.create(:devices, @device2)
+              @adapter.create(:devices, @device3)
+
+              TestUserRepository.adapter = @adapter
+            end
+
+            it 'returns records' do
+              created_user = TestUserRepository.first
```
jodosha: Same as: https://github.com/lotus/model/pull/228/files#r39265171

jodosha:
@AlfonsoUceda It looks good! :clap: There are a couple of inline comments I left here and Ruby warnings when the tests are running.

jodosha:
Lovely work, merging it! :clap:

leandrocp:
:beers: great work !

PR #230, opened 2015-08-18
Adds simple BCrypt secure passwords

benlovell:
Includes basic validation for password and confirmation. Ensures
passwords are kept below 72 characters, the theoretical maximum
permitted by bcrypt's hashing function.

I hit some troubles with Lotus validations. It was a surprise for me to
find that the validator macro creates accessors/attributes. It would
have been tidier to include a module containing the instance methods
within the `included` callback, but due to ordering with the attributes
API, and my need to overload the `password=` accessor _after_ it had
been defined by the attributes API.

Because I wanted to use the confirmation, and the `size` validators
this pushed me toward that particular approach.

Some things to note:

* BCrypt is latently loaded and needs to be present in the Gemfile,
 this will be documented of course.
* No efforts were made to abstract access to BCrypt since it's just the
 defacto approach/library. Happy to do this though if anyone feels
 strongly.
* At the moment, the hashing cost is whatever BCrypt's default is. This
 can be overridden via `BCrypt::Engine.cost=`. It may be worth
 providing a nicer API here for consumers.

I just wanted to get some validation on the approach before I
documented the code. WDYT?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,8 @@ end
 gem 'lotus-utils',       '~> 0.5', require: false, github: 'lotus/utils',       branch: '0.5.x'
 gem 'lotus-validations',           require: false, github: 'lotus/validations', branch: '0.3.x'
 
+gem 'bcrypt', require: false
+
 platforms :ruby do
   gem 'sqlite3', require: false
   gem 'pg'
--- a/lib/lotus/entity/secure_password.rb
+++ b/lib/lotus/entity/secure_password.rb
@@ -1,0 +1,39 @@
+require 'lotus/validations'
+
+module Lotus
+  module Entity
+    module SecurePassword
+      # Maximum as permitted by bcrypt's hash function
+      MAX_LENGTH = 72
+
+      def self.included(base)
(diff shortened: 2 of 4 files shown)


jodosha:
@benlovell Thanks for this PR. It's a great addition for devs' productivity. :smile: :+1: But, I don't see it to fit well in the context of an ORM. Maybe this is a chance to create the first plugin for Lotus Model. What do you think? :green_heart: :yellow_heart: :blue_heart:

benlovell:
@jodosha No problem, that makes sense to me. I'll extract this into a gem. :heart_decoration:

AlfonsoUceda:
@benlovell tell us when you create the gem. we have create a lotus gems section, thanks!

LeoNogueira:
Hi all!

I used @benlovell code and packaged a simple gem at [](https://github.com/leonogueira/secure_password).

That's my first attempt to contribute to OSS and Lotus in particular. 

I was not sure what namespace should be used to consider the gem a 'plugin'.

I would love you feedback and hope I could help.

:gift_heart:

PR #235, opened 2015-09-03
Drop support for mysql adapter in favour of mysql2 adapter

joneslee85:
Sequel supports both mysql and mysql2, in majority of cases mysql2 are opted to mysql. By adopting one mysql engine, I think it would reduce the maintenance for the team and enhances the consistency

### DIFF
--- a/lib/lotus/model/adapters/sql_adapter.rb
+++ b/lib/lotus/model/adapters/sql_adapter.rb
@@ -43,9 +43,8 @@ module Lotus
         # @since 0.1.0
         def initialize(mapper, uri)
           super
-          @connection = Sequel.connect(@uri)
-        rescue Sequel::AdapterNotFound => e
-          raise DatabaseAdapterNotFound.new(e.message)
+          _preprocess_uri
+          connect
         end
 
         # Creates a record in the database for the given entity.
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
Why this change? is it better mysql2 than mysql?

to me :+1:

brennovich:
@AlfonsoUceda yep, it is. `mysql` is outdated and I don't even know if it run against Ruby 2.2 http://stackoverflow.com/questions/5411551/what-the-difference-between-mysql-and-mysql2-gem

jodosha:
@joneslee85 While `mysql2` is better than `mysql` this is true only for MRI. With the goal in mind to support JRuby, the jdbc adapter doesn't recognize the `mysql2://` scheme for URL and raises an exception:

```
  1) Error:
Lotus::Model::Adapters::SqlAdapter::#initialize#test_0004_changes the URI of mysql:// to mysql2://:
Lotus::Model::Adapters::DatabaseAdapterNotFound: LoadError: no such file to load -- mysql2
    /Users/luca/Code/lotus/model/lib/lotus/model/adapters/sql_adapter.rb:264:in `connect'
    /Users/luca/Code/lotus/model/lib/lotus/model/adapters/sql_adapter.rb:47:in `initialize'
    /Users/luca/Code/lotus/model/test/model/adapters/sql_adapter_test.rb:83:in `block in test_0004_changes the URI of mysql:// to mysql2://'
```

I'm gonna to revert this change for JRuby compat and to leave freedom to devs to use the low level adapter they need. That being said, `mysql2` must be the one that we promote, like we did with #331. Good job :clap:

brennovich:
@jodosha please don't revert it. We need to treat as a special case, because neither `mysql` nor `mysql` works for jruby, they are MRI only. We need to ensure that `jdbc-mysql` will be used for jruby

jodosha:
@brennovich Don't you think it's up to the devs to provide the right URI in their `.env` files?

brennovich:
You're are right, what I said in my previous comment applies only for lotus generators, I think ;)

joneslee85:
@jodosha fair point, please revert it

jodosha:
@joneslee85 @brennovich I reverted and removed GH milestone.

Issue #237, opened 2015-09-12
Repository does not respect default values on persisted entities


taylorfinnell:
When you define a default value in the migrator, the repository does not respect it.

```
Lotus::Model.migration do
  up do
    create_table :user_info do
      primary_key :id

      column :rank, Integer, default: -1
      ...
```

`ui = UserInfoRepository.create(UserInfo.new(version: 1))`

```
ui.rank
 => nil
```

joneslee85:
@taylorfinnell thanks for the bug report, for our investigation, can you please also provide your mapper?

taylorfinnell:
Mapping

```
collection :user_info do
  entity      Fastpeek::UserInfo
  repository  Fastpeek::UserInfoRepository

  attribute :id,   Integer
  attribute :rank, Integer
end
```

Using the SQL adapter. Here is a potential solution for the issue, originally used the schema for the defaults but @solnic showed me the Dataset#returning method :)

https://github.com/taylorfinnell/model/commit/52f7683d88308fb8db265d926194c8ebab90e5ef

taylorfinnell:
Got a chance to run the tests. Looks like sqlite does not support RETURNING. So the commit in my previous comment is not a fully working solution

PR #239, opened 2015-09-15
respect default entity values

taylorfinnell:
See #237

### DIFF
--- a/lib/lotus/model/adapters/sql/collection.rb
+++ b/lib/lotus/model/adapters/sql/collection.rb
@@ -55,12 +55,11 @@ module Lotus
           # @api private
           # @since 0.1.0
           def insert(entity, defaults)
+            _apply_defaults(entity, defaults)
+
             serialized_entity            = _serialize(entity)
             serialized_entity[identity] = super(serialized_entity)
 
-            defaults.each_pair do |attribute, default|
-              serialized_entity[attribute] ||= default
-            end
 
             _deserialize(serialized_entity)
           end
@@ -283,6 +282,14 @@ module Lotus
           # @since 0.2.2
           def _deserialize(entity)
             @mapped_collection.deserialize([entity]).first
+          end
+
+          def _apply_defaults(entity, defaults)
+            defaults.each_pair do |attribute, default|
+              if entity.send(attribute).nil?
+                entity.send("#{attribute}=", default)
+              end
+            end
           end
         end
       end
--- a/lib/lotus/model/adapters/sql/collection.rb
+++ b/lib/lotus/model/adapters/sql/collection.rb
@@ -287,7 +287,7 @@ module Lotus
           def _apply_defaults(entity, defaults)
             defaults.each_pair do |attribute, default|
               if entity.send(attribute).nil?
-                entity.send("#{attribute}=", default)
+                entity.send("#{attribute}=", @mapped_collection.deserialize_attribute(attribute, default))
               end
             end
           end
--- a/lib/lotus/model/adapters/sql/collection.rb
+++ b/lib/lotus/model/adapters/sql/collection.rb
@@ -54,9 +54,13 @@ module Lotus
           #
           # @api private
           # @since 0.1.0
-          def insert(entity)
+          def insert(entity, defaults)
             serialized_entity            = _serialize(entity)
             serialized_entity[identity] = super(serialized_entity)
+
+            defaults.each_pair do |attribute, default|
+              serialized_entity[attribute] ||= default
+            end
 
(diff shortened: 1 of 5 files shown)


pascalbetz:
@taylorfinnell why not checking for the presence of the key instead of checking for a _trueish_ value? 
The question i'd ask is "has there been a default assigned" not "is there a default which is true".
Same problem as with assigning the the defaults, actually.

```ruby
if properties[:ruby_default] || properties[:db_type] == 'boolean'
```

=> 

```ruby
if properties.key?(:ruby_default)
```

lib/lotus/model/adapters/sql_adapter.rb
```diff
@@ -281,6 +281,28 @@ def disconnect
 
         private
 
+        # Returns a hash of attributes and their default values
+        #
+        # @api private
+        # @since x.x.x
+        def defaults(collection)
```
jodosha: This logic can be a private method of `Lotus::Model::Adapters::Sql::Collection`. This is a delegator around a Sequel dataset, so you have direct access to schema informations (via `#schema`).

jodosha:
@taylorfinnell Thanks for starting this.

I think this should honor the mapper first and then the database schema defaults. As of `v0.5.0`, we'll have custom coercers (see #216). That means, a dev can setup their own serialize/deserialize policy for an attribute. I'll give you one silly example to explain the concept.

Let's say we have a `price` column expressed as an integer with cents, so `1€` is stored as `100`. In the Ruby world, the dev wants a float for that attribute, so `1€` is `1.00`. Here's the corresponding code.

```ruby
# migration
column :price, Integer, default: 100
```

```ruby
class Price
  def self.dump(value)
    Integer(value) unless value.nil?
  end

  def self.load(value)
    Float(value) unless value.nil?
  end
end

# mapper
attribute :price, Price
```

Now, with this current implementation, we have a problem:

```ruby
book = Book.new
book = BookRepository.create(book)
book.price # => 100 WRONG
```

We wanted a float, but we have an integer. To solve this issue, we should use `#deserialize_attribute`.

/cc @lotus/core-team @pascalbetz

joneslee85:
@taylorfinnell ping

taylorfinnell:
Oops! Was not getting notifications for this, will work on it this week! Thanks guys

DanielBlanco:
Is there an ETA for this fix? right now I'm doing this:
```
foo = FooRepository.create(foo)
foo = FooRepository.find(foo.id)  # <= Shouldn't need to do this
```
I need it to get default `Sequel.function(:uuid_generate_v4)` value

hlegius:
Guys, any news about it?

jodosha:
@hlegius Nobody's working on it at the moment.

joneslee85:
ping

pascalbetz:
@taylorfinnell posted a question on gitter:

@jodosha Do you have any tips on how to handle the issue mentioned in the first comment here. taylorfinnell/model@0759f90. Working on #237 (lotus/model#239) and have have started to use the deserialize_attribute like you mention but I am not sure how to tell if the user has set their value to nil or if the value is nil and because no value was set and it needs a default

lib/lotus/model/adapters/sql_adapter.rb
taylorfinnell: It looks like #schema is actually on the Sequel connection object not the Sequel dataset object

taylorfinnell:
It looks like attributes that are set to nil don't make it to #create.

Fake migration...
```
  create_table :users do
     column :name, String, default: "Jim"
end
```

```
u = TestUser.new(name: nil)
UserRepo.create(u)
```

In Sql::Collection#insert the `serialized_entity` at that point only contains `id` and no `name`. So I think it would seem reasonable to set the defaults at that point like in the commit @pascalbetz mentioned above

lib/lotus/model/adapters/sql/collection.rb
```diff
@@ -280,6 +283,14 @@ def _serialize(entity)
           def _deserialize(entity)
             @mapped_collection.deserialize([entity]).first
           end
+
+          def _apply_defaults(entity, defaults)
+            defaults.each_pair do |attribute, default|
+              if entity.send(attribute).nil?
```
taylorfinnell: Maybe it's better to serialize the entity and apply the defaults to the hash and then deserialize

pascalbetz: if you serialize first, then you could probably avoid the ```send```?

jodosha:
@taylorfinnell I'm closing this in favor of a new implementation that I'm working on. Thanks!

PR #247, opened 2015-10-12
Return error_message for dropdb messages

joneslee85:
I made a mistake with this work in the past. The return message should be the `error_message`
cc @lotus/core-team


AlfonsoUceda:
can we test this?

joneslee85:
@AlfonsoUceda we can, I'll see what I can

PR #253, opened 2015-10-14
refactor check attribute name

aderyabin:
Currently `defined_attribute?` has strange realization and usage. It checks name on inclusion in `instance_methods`.
Next point that attrubute wil be added to attribute list even if `defined_attribute?(attr)` returns `false`

I suggest small fixes: 
– rename to `allowed_attribute_name?`. Method name is clear for understanding
– add to attributes only allowed by name instead of all. It helps to avoid duplicated attributes too.
– add tests on this logic

### DIFF
--- a/test/entity_test.rb
+++ b/test/entity_test.rb
@@ -38,11 +38,31 @@ describe Lotus::Entity do
       Car.attributes.must_equal Set.new([:id, :model])
     end
 
+    it 'rejects existed instance methods' do
+      Car.attributes :object_id
+      Car.attributes.must_equal Set.new([:id])
+    end
+
     describe 'params is array' do
       it 'defines attributes' do
         Car.attributes [:model]
         Car.attributes.must_equal Set.new([:id, :model])
       end
+    end
+  end
+
+  describe '.allowed_attribute_name?' do
+    it 'returns true if attrubute not defined' do
+      Car.allowed_attribute_name?(:model).must_equal true
+    end
+
+    it 'returns false if attribute defined' do
+      Car.attributes :model
+      Car.allowed_attribute_name?(:model).must_equal false
+    end
+
+    it 'returns false for reserved word' do
+      Car.allowed_attribute_name?(:to_json).must_equal false
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/entity_test.rb
```diff
@@ -38,6 +38,11 @@ class Camera
       Car.attributes.must_equal Set.new([:id, :model])
     end
 
+    it 'reject existed instance methods' do
```
joneslee85: s/reject/rejects

test/entity_test.rb
```diff
@@ -46,6 +51,25 @@ class Camera
     end
   end
 
+  describe '.allowed_attribute_name?' do
+    it 'return true if attrubute not defined' do
+      Car.allowed_attribute_name?(:model).must_equal true
+    end
+
+    it 'return false if attribute defined' do
```
joneslee85: s/return/returns

test/entity_test.rb
```diff
@@ -46,6 +51,25 @@ class Camera
     end
   end
 
+  describe '.allowed_attribute_name?' do
+    it 'return true if attrubute not defined' do
```
joneslee85: s/return/returns

test/entity_test.rb
```diff
@@ -46,6 +51,25 @@ class Camera
...
+    it 'return false if attribute defined' do
+      Car.attributes :model
+      Car.allowed_attribute_name?(:model).must_equal false
+    end
+
+    it 'id is not allowed by default' do
+      Car.allowed_attribute_name?(:id).must_equal false
+    end
+
+    it 'return false for reserved word' do
```
joneslee85: s/return/returns

test/entity_test.rb
```diff
@@ -46,6 +51,25 @@ class Camera
...
+    it 'return true if attrubute not defined' do
+      Car.allowed_attribute_name?(:model).must_equal true
+    end
+
+    it 'return false if attribute defined' do
+      Car.attributes :model
+      Car.allowed_attribute_name?(:model).must_equal false
+    end
+
+    it 'id is not allowed by default' do
```
joneslee85: can we rephrase this description to a verb?

PR #255, opened 2015-10-15
Entity identity

aderyabin:
I suggest to have entity identity configurable instead of hard-coced `:id`. In some cases it may be `uuid` for example

```ruby
require 'lotus/model'
class User
  include Lotus::Entity
  identity :uuid
end
```
It compatible with existed models because default entity identity is `:id`.

### DIFF
--- a/test/entity_test.rb
+++ b/test/entity_test.rb
@@ -343,4 +343,64 @@ describe Lotus::Entity do
       end
     end
   end
+
+  describe '#identity' do
+    describe 'default behavior' do
+      it 'sets :id' do
+        Car.identity.must_equal :id
+      end
+
+      describe ".attributes" do
+        it 'defines :identity by default' do
+          Car.attributes.must_equal Set.new([:id])
+        end
+      end
+
+      describe "#id" do
+        it 'is accessible' do
+          car = Car.new
+          car.id = 100
+          car.id.must_equal 100
+        end
+      end
+    end
+
+    describe 'new value' do
+      before do
+        class Post
+          include Lotus::Entity
+          identity :uuid
+
+          attributes :name
+        end
+      end
+
+      it 'sets new identity' do
+        Post.identity :uuid
+        Post.identity.must_equal :uuid
+      end
+
+      it 'replaces old identity' do
+        Post.attributes.must_equal Set.new([:uuid, :name])
+        post = Post.new
+        proc{post.id}.must_raise NoMethodError
+        proc{post.id = 100}.must_raise NoMethodError
+      end
+
+      it 'sets new attribute' do
+        Post.attributes.must_equal Set.new([:uuid, :name])
+        post = Post.new
+        post.uuid = 100
+        post.uuid.must_equal 100
+      end
+    end
+  end
+
+  describe ".identity" do
+    let(:book) { Book.new(id: 23, title: 'Wuthering Meadow', author: 'J. K. Rowling', published: true ) }
+
+    it 'returns identity value' do
+      book.identity.must_equal 23
+    end
+  end
 end
--- a/lib/lotus/entity.rb
+++ b/lib/lotus/entity.rb
@@ -81,7 +81,7 @@ module Lotus
     def self.included(base)
       base.class_eval do
         extend ClassMethods
-        attributes :id
+        attributes identity
       end
     end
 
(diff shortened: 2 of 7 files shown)


test/entity_test.rb
```diff
@@ -320,4 +320,61 @@ class Camera
...
+    describe "instance" do
+      let(:book) { Book.new(id: 23, title: 'Wuthering Meadow', author: 'J. K. Rowling', published: true ) }
+
+      it 'returns identity value' do
+        book.identity.must_equal 23
+      end
+    end
+
+    describe 'default behavior' do
+      it 'equals :id' do
```
joneslee85: s/equals/returns

test/entity_test.rb
```diff
@@ -320,4 +320,61 @@ class Camera
...
+
+    describe "instance" do
+      let(:book) { Book.new(id: 23, title: 'Wuthering Meadow', author: 'J. K. Rowling', published: true ) }
+
+      it 'returns identity value' do
+        book.identity.must_equal 23
+      end
+    end
+
+    describe 'default behavior' do
```
joneslee85: this should be in a separate `describe '.identity'`

test/entity_test.rb
```diff
@@ -320,4 +320,61 @@ class Camera
...
+      end
+
+      it 'replaces old identity' do
+        Post.attributes.must_equal Set.new([:uuid, :name])
+        post = Post.new
+        proc{post.id}.must_raise NoMethodError
+        proc{post.id = 100}.must_raise NoMethodError
+      end
+
+      it 'sets new attrubute' do
```
joneslee85: s/attrubute/attribute

test/entity_test.rb
```diff
@@ -320,4 +320,61 @@ class Camera
...
+      before do
+        class Post
+          include Lotus::Entity
+          identity :uuid
+
+          attributes :name
+        end
+      end
+
+      it 'sets new' do
```
joneslee85: s/sets new/sets new identity

test/entity_test.rb
```diff
@@ -320,4 +320,61 @@ class Camera
...
+        book.identity.must_equal 23
+      end
+    end
+
+    describe 'default behavior' do
+      it 'equals :id' do
+        Car.identity.must_equal :id
+      end
+
+      it ':id included in attributes' do
```
joneslee85: what do you think about grouping this under `#attributes` instead?

test/entity_test.rb
```diff
@@ -320,4 +320,61 @@ class Camera
...
+        Car.attributes.must_equal Set.new([:id])
+      end
+
+      it ':id is accessible' do
```
joneslee85: same goes here, what about grouping under `#id`

joneslee85:
generally I like this PR, please rebase

cc @lotus/core-team

AlfonsoUceda:
@aderyabin I thin it's a little weird this implementation, instead add `attribute` in included method, have you tried to add the identity in the initialize method? maybe it's simpler than current implementation

aderyabin:
@AlfonsoUceda If I correct understand you, you are talking about 
```
    def self.included(base)
      base.class_eval do
        extend ClassMethods
        attributes identity
      end
    end
```
and replacing `attributes identity` to '#initialize'.

In this case `User.attributes` will be without identity attribute defined on `initialize` later.

AlfonsoUceda:
You are right, forgot then

jodosha:
@aderyabin Hello again. Thanks for all these discussions and contributions. :smile: 

Is `uuid` coming from your database? If so, would this solve your problem?

```ruby
mapping do
  collection :users do
    # ...
    attribute :id, Integer, as: :uuid
  end
end
```

aderyabin:
Yes, that is a solution. But I don't like idea to have hardcoded `.id` usage.

aderyabin:
Yes, I understand that my suggestion is ambiguous and usage will be do not widespread.

AlfonsoUceda:
I think we should provide a way to set a custom identity for an entity with the correct attribute and don't use always id

jodosha:
@aderyabin @AlfonsoUceda Because we have the mapper, there is no need to configure the identity. By sticking to `id` helps to keep the internals of `Entity` simple.

PR #258, opened 2015-10-19
Id flexibility

aderyabin:
The PR is a result of our discussion with @jodosha in #248 .

I removed `.persist` and check on `id` in `.create`. It give a chance to set `id` manually. Related test also included.

```ruby
@persisted_user = UserRepository.create(User.new(name: 'My', age: '23', id: 100))
UserRepository.all # => [@persisted_user]
@persisted_user.id #=> 100
```

### DIFF
--- a/lib/lotus/model.rb
+++ b/lib/lotus/model.rb
@@ -17,6 +17,21 @@ module Lotus
     #
     # @see Lotus::Repository.update
     class NonPersistedEntityError < ::StandardError
+      def initialize(message = "Entity is not persisted")
+        super
+      end
+    end
+
+    # Error for already persisted entity
+    # It's raised when we try to create already persisted entry.
+    #
+    # @since 0.6.0
+    #
+    # @see Lotus::Repository.create
+    class AlreadyPersistedEntityError < ::StandardError
+      def initialize(message = "Entity is already persisted")
+        super
+      end
     end
 
     # Error for invalid mapper configuration
--- a/lib/lotus/model/adapters/memory/collection.rb
+++ b/lib/lotus/model/adapters/memory/collection.rb
@@ -32,6 +32,29 @@ module Lotus
               yield(@current += 1)
               @current
             end
+
+            # Sets the current count for a given value
+            #
+            # @param id [Fixnum] the value to set
+            #
+            # @see Lotus::Model::Adapters::Memory::Command#create
+            #
+            # @return [Fixnum] the counter
+            #
+            # @api private
+            # @since 0.6.0
+            def set!(id)
+              @current = id
+            end
+          end
+
+          # Error for identity unique constraint
+          # It's raised when new entity duplicates identity value
+          #
+          # @since 0.6.0
+          #
+          # @see Lotus::Model::Adapters::Memory::Collection#create
+          class UniqueConstraintViolation < ::StandardError
           end
 
           # @attr_reader name [Symbol] the name of the collection (eg. `:users`)
@@ -77,9 +100,10 @@ module Lotus
           # @api private
           # @since 0.1.0
           def create(entity)
-            @primary_key.increment! do |id|
-              entity[identity] = id
-              records[id] = entity
+            if entity[identity]
+              create_with_id(entity)
+            else
+              create_without_id(entity)
             end
           end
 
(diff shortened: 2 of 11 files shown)


lib/lotus/model/adapters/memory/collection.rb
```diff
@@ -32,6 +32,13 @@ def increment!
               yield(@current += 1)
               @current
             end
+
```
joneslee85: can we document this method too?

lib/lotus/model/adapters/memory/collection.rb
```diff
@@ -32,6 +32,13 @@ def increment!
...
+            def set!(id)
+              @current = id
+            end
+          end
+
+          class UniqueConstraintViolation < ::StandardError
```
joneslee85: please document this class

lib/lotus/model/adapters/memory/collection.rb
```diff
@@ -77,9 +84,18 @@ def initialize(name, identity)
           # @api private
           # @since 0.1.0
           def create(entity)
-            @primary_key.increment! do |id|
-              entity[identity] = id
-              records[id] = entity
+            if _id = entity[identity]
+              if records[_id]
+                raise UniqueConstraintViolation
```
joneslee85: what do you think about adding messages to help user identify which entity fail this constraint?

aderyabin:
@joneslee85 Done!

lib/lotus/repository.rb
```diff
@@ -296,10 +249,14 @@ def persist(entity)
...
       #   created_article # => nil
       #
+      #   imported_article = ArticleRepository.create(Article.new(title: 'New Lotus features', id: 123))
+      #   imported_article.id  # => 123
       def create(entity)
-        unless _persisted?(entity)
+        begin
           _touch(entity)
           @adapter.create(collection, entity)
+        rescue Sequel::UniqueConstraintViolation, Lotus::Model::Adapters::Memory::Collection::UniqueConstraintViolation => e
```
AlfonsoUceda: I think we don't have to rescue an exception here, this should be handled by the user, IMO

lib/lotus/model/adapters/memory/collection.rb
```diff
@@ -77,9 +100,18 @@ def initialize(name, identity)
           # @api private
           # @since 0.1.0
           def create(entity)
-            @primary_key.increment! do |id|
-              entity[identity] = id
-              records[id] = entity
+            if _id = entity[identity]
+              if records[_id]
+                raise UniqueConstraintViolation, "Duplicate identity value violates unique constraint"
+              else
```
AlfonsoUceda: I think you can remove this else:

```ruby
if records[_id]
  raise UniqueConstraintViolation, "Duplicate identity value violates unique constraint"
end
records[_id] = entity
@primary_key.set!(records.keys.max)
```

AlfonsoUceda: what do you think to create to private methods to give context?

```ruby
def create(entity)
  if entity[identity]
    create_with_id(entity)
  else
    create_without_id(entity)
  end
end
```

lib/lotus/repository.rb
jodosha: Totally agree here, we are letting the operation to silently fail. This shouldn't be caught.

lib/lotus/model.rb
```diff
@@ -17,6 +17,9 @@ module Model
     #
     # @see Lotus::Repository.update
     class NonPersistedEntityError < ::StandardError
+      def initialize(message = "Entry is not persisted")
```
jodosha: Entity, instead of Entry :wink:

test/repository_test.rb
```diff
@@ -170,9 +73,23 @@
             UserRepository.delete(@persisted_user)
           end
 
-          it 'does not touch created_at' do
-            UserRepository.create(@persisted_user)
-            @persisted_user.created_at.wont_be_nil
+          it 'raises an exception' do
+            -> { UserRepository.create(@persisted_user) }.must_raise Exception
```
jodosha: @aderyabin `Exception` is generic, can you please test against the specific type? Thanks!

jodosha:
@aderyabin This looks good, can you please rebase in order to merge it? Thank you! :+1:

aderyabin:
@jodosha review it again, please

joneslee85:
please rebase

aderyabin:
Done!

pascalbetz:
Why _persist_ has been removed? It's quite useful, i think.

lib/lotus/repository.rb
```diff
@@ -296,10 +249,14 @@ def persist(entity)
...
+      #   imported_article = ArticleRepository.create(Article.new(title: 'New Lotus features', id: 123))
+      #   imported_article.id  # => 123
       def create(entity)
-        unless _persisted?(entity)
-          _touch(entity)
+        _touch(entity)
+        begin
           @adapter.create(collection, entity)
+        rescue Exception => e
+          raise Lotus::Model::AlreadyPersistedEntityError.new e.message
```
pascalbetz: is this the right error type in all cases where an _Exception_ can occur from the underlying ```@adapter```?

jodosha: Yeah, @pascalbetz is right:

1. [Never catch `Exception`](http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby). 
2. There are exceptions that we want to let to bubble, for instance a database integrity violation.
3. It's too generic to assume that all the exceptions are raised because the entity was already persisted.

hlegius:
Any update in this task? /cc @jodosha

jodosha:
@hlegius We want to freeze all the changes for Model, because we want to better understand if the current framework design can be improved.

jodosha:
@aderyabin Closing this in favor of a new implementation I'm working on. Thanks!

Issue #257, opened 2015-10-19
Improve tests


aderyabin:
I want to create PR that add possibility to run test with specific adapter, for example
```
rake test:postgresql
```

Does it sense?

// cc @jodosha @joneslee85

joneslee85:
@aderyabin can't we do it with UNIX variable instead?

aderyabin:
yes, we can do it like in AR
```
ARCONN=mysql rake test
```

joneslee85:
I like it, but wondering if it would brings the mutations of Travis CI matrix to unmanageable number

jodosha:
> I like it, but wondering if it would brings the mutations of Travis CI matrix to unmanageable number

What if the default runs all the tests (like now), so we can use it with Travis CI, without growing the build matrix? I see this trick useful for dev mode. @aderyabin was this your intention?

aderyabin:
I suggest only add option – run test with specific adapter, for example:
```
ADAPTER=postgres rake test
``` 
without changes in default task (test all adapters)
```
rake test
```

jodosha:
@aderyabin :+1:

joneslee85:
:+1:

hlegius:
@aderyabin I was talking to @jodosha about to split tests into two groups: unit and integration. IMO, it will help to prevent 'too many open files' error in the future due lots of resources left behind without closing it properly (sometimes its too way complicated to close anyway). Also, may improve _our_ development process. This separation will be also reflected within Travis which will call explicitly unit and integration in this order of calling to avoid the first item mentioned above. @jodosha agreed (at the first time :P). What do you think, guys?

pascalbetz:
i agree. but i don't have to manage the build matrix at travis:-)

but i disagree that this should be the fix to the "too many open files" problem. If this exists, then it's a problem with the tests allocating to many resources and should be fixed. Currently we assume that it's the tests which are causing these problems. But it might as well be the framework itself. Which would then go unnoticed.

hlegius:
@pascalbetz hmm.. I've fixed the 'too many open files' issue days ago. it was merged into master one day later. My concern is about when new files get added into repository people forget to close connections properly (as it was the issue before) - that's why I suggested split within Travis as well.

joneslee85:
generally I find this idea useful for dev mode. Furthermore I think I would make tests simpler and easier to track down bugs.

Please go ahead

jodosha:
> generally I find this idea useful for dev mode.

[...]

PR #261, opened 2015-10-27
Added test:unit and test:integration

hlegius:
I've split tests into two categories: unit and integration. This way we can run our unit suite on dev mode faster leaving integration test to a phase pre commit or even leaving it with Travis, for example.

To get it done, I just needed to move one file from `test/` to `test/integration`. I left `rake test` as was before: running both kind of tests.

I believe this PR won't bring any new complexity to @aderyabin idea to run tests by db adapters. 

`rake test:unit`
> 498 runs, 664 assertions

`rake test:integration`
> 252 runs, 744 assertions

`rake test`
> 750 runs, 1408 assertions

IMO, would be nice to modify Travis to read both: unit and integration, respectively - but it's up to you, guys :+1: 

Refers to #257 (this won't close the entire issue; just deliver 50% of it.)
/cc @jodosha @aderyabin


jodosha:
@hlegius Thank you :+1:

PR #263, opened 2015-10-30
Allow an after_connect callback for repo adapters

taylorfinnell:
I needed to be able to access the underlying connection object from Sequel in order to turn on auto commit for a jdbc connection. Since Sequel has an `after_connect` callback I tried to expose it to lotus-model in a reasonable way as well as provide support for the same callback to other repositories for consistency.  Some notes about `after_connect` and its uses: http://sequel.jeremyevans.net/rdoc/files/doc/release_notes/3_11_0_txt.html

My usecase looks like:

```
Lotus::Model.configure do
  adapter type: :sql, uri: 'jdbc:...', after_connect: Proc.new { |connection| connection.auto_commit(true) }

  mapping do
  end
end
```

Please let me know your thoughts about the idea :)

### DIFF
--- a/lib/lotus/model/adapters/abstract.rb
+++ b/lib/lotus/model/adapters/abstract.rb
@@ -85,6 +85,8 @@ module Lotus
         #
         # @param uri [String] the optional connection string to the database
         #
+        # @param options [Hash] a list of non-mandatory adapter options
+        #
         # @since 0.1.0
         def initialize(mapper, uri = nil, options = {})
           @mapper = mapper
--- a/lib/lotus/model/adapters/abstract.rb
+++ b/lib/lotus/model/adapters/abstract.rb
@@ -86,9 +86,10 @@ module Lotus
         # @param uri [String] the optional connection string to the database
         #
         # @since 0.1.0
-        def initialize(mapper, uri = nil)
+        def initialize(mapper, uri = nil, callbacks = {})
           @mapper = mapper
           @uri    = uri
+          @callbacks = callbacks
         end
 
         # Creates or updates a record in the database for the given entity.
--- a/lib/lotus/model/adapters/abstract.rb
+++ b/lib/lotus/model/adapters/abstract.rb
@@ -86,10 +86,10 @@ module Lotus
         # @param uri [String] the optional connection string to the database
         #
         # @since 0.1.0
-        def initialize(mapper, uri = nil, callbacks = {})
+        def initialize(mapper, uri = nil, options = {})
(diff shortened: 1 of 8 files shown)


jodosha:
@taylorfinnell Hello and thanks for this PR.

Sorry, but I think this approach is too specific. What I would like to see here is a generic set of non-mandatory options to pass down to the adapter. It's then responsible to handle or ignore them.

This opens a world of possibilities larger than callbacks. Eg. if you need to pass a `foo:` option to Sequel, you can do with generic options, but not with the approach above.

Can you please remove the notion of callbacks from the PR and pass down options to the raw adapter (eg `Sequel`)? Once done, I'll be more than happy to merge this. Thank you :+1:

taylorfinnell:
@jodosha Thank you for the input! I have made the changes how I believe you want. Please let me know! I am wondering about the MemoryAdapter, currently the options are just ignored. Should I maybe throw an exception about options not being supported for MemoryAdapter, maybe a warning?

lib/lotus/model/adapters/file_system_adapter.rb
```diff
@@ -66,7 +66,7 @@ class FileSystemAdapter < MemoryAdapter
         #
         # @api private
         # @since 0.2.0
-        def initialize(mapper, uri)
+        def initialize(mapper, uri, callbacks = {})
```
jodosha: @taylorfinnell Please change `callbacks` into `options`.

lib/lotus/model/adapters/abstract.rb
```diff
@@ -86,9 +86,10 @@ class Abstract
         # @param uri [String] the optional connection string to the database
         #
         # @since 0.1.0
-        def initialize(mapper, uri = nil)
+        def initialize(mapper, uri = nil, options = {})
```
jodosha: @taylorfinnell Can you please document `options` as `@param` in the API doc? Don't forget the other signatures! :smile: Thank you :+1:

taylorfinnell:
Updated the variable name, added the docs. Did you mean that I should document all the available options? I suppose the option documentation would go on the adapters themselves. SqlAdapter would note its available options (after_connect, max_connections, etc), MemoryAdapter would note its available options (none)

jodosha:
@taylorfinnell What I mean, is to add `@param options [Hash] description here..`

taylorfinnell:
@jodosha I think that is all done :)

jodosha:
@taylorfinnell It looks good now! Merging :+1:

PR #265, opened 2015-11-29
Catch SystemCallError, in case command doesn't exist

### LINKED ISSUE #264: Show error message for `createdb` failure
I'm using [Postgres.app](http://postgresapp.com/) on OS X, and I just ran into a problem running `lotus db create`

```
 ~> lotus db create
/Users/sean/.gem/ruby/2.1.6/gems/lotus-model-0.5.0/lib/lotus/model/migrator/postgres_adapter.rb:29:in ``': No such file or directory - createdb (Errno::ENOENT)
	from /Users/sean/.gem/ruby/2.1.6/gems/lotus-model-0.5.0/lib/lotus/model/migrator/postgres_adapter.rb:29:in `create'
	from /Users/sean/.gem/ruby/2.1.6/gems/lotus-model-0.5.0/lib/lotus/model/migrator.rb:84:in `create'
	from /Users/sean/.gem/ruby/2.1.6/gems/lotusrb-0.5.0/lib/lotus/commands/db/create.rb:9:in `start'
	from /Users/sean/.gem/ruby/2.1.6/gems/lotusrb-0.5.0/lib/lotus/commands/db.rb:31:in `create'
	from /Users/sean/.gem/ruby/2.1.6/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /Users/sean/.gem/ruby/2.1.6/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/sean/.gem/ruby/2.1.6/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /Users/sean/.gem/ruby/2.1.6/gems/thor-0.19.1/lib/thor/invocation.rb:115:in `invoke'
	from /Users/sean/.gem/ruby/2.1.6/gems/thor-0.19.1/lib/thor.rb:235:in `block in subcommand'
```

The problem was that running the `createdb ...` on my computer didn't succeed. I needed to follow [these directions](http://postgresapp.com/documentation/cli-tools.html) (i.e. add `createdb` to my PATH) in order to make it work.

Related: #200, #262

I solved it by looking at `postgres_adapter.rb`, but should we see if `createdb` fails, and show an error saying 'running `createdb` failed` instead of the above?

cllns:
Closes #264.

I don't really like that we basically just change the `SystemCallError` into a `MigrationError`, without rescuing it, but that seems like another PR :)

### DIFF
--- a/test/integration/migration/postgres_test.rb
+++ b/test/integration/migration/postgres_test.rb
@@ -71,6 +71,23 @@ describe 'PostgreSQL Database migrations' do
         exception.message.must_include 'dropdb: database removal failed'
         exception.message.must_include 'There is 1 other session using the database'
       end
+
+      describe "when a command isn't available" do
+        before do
+          # We accomplish having a command not be available by setting PATH
+          # to an empty string, which means *no commands* are available.
+          @original_path = ENV['PATH']
+          ENV['PATH'] = ''
+        end
+
+        it "raises MigrationError on create" do
+          -> { Lotus::Model::Migrator.create }.must_raise Lotus::Model::MigrationError
+        end
+
+        after do
+          ENV['PATH'] = @original_path
+        end
+      end
     end
 
     describe "migrate" do
--- a/lib/lotus/model/migrator/postgres_adapter.rb
+++ b/lib/lotus/model/migrator/postgres_adapter.rb
@@ -103,12 +103,14 @@ module Lotus
         def call_db_command(command)
           require 'open3'
 
-          Open3.popen3(command, database) do |stdin, stdout, stderr, wait_thr|
-            exit_status = wait_thr.value
-
-            unless exit_status.success?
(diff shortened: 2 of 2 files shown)


test/integration/migration/postgres_test.rb
```diff
@@ -71,6 +71,23 @@
...
+      describe "when a command isn't available" do
+        before do
+          # We accomplish having a command not be available by setting PATH
+          # to an empty string, which means *no commands* are available.
+          @original_path = ENV['PATH']
+          ENV['PATH'] = ''
+        end
+
+        it "raises MigrationError on create" do
+          -> { Lotus::Model::Migrator.create }.must_raise Lotus::Model::MigrationError
```
joneslee85: can we asset the exception message too?

PR #266, opened 2015-12-03
Introduce Lotus::Model::Error

karimmtarek:
- Added Error class
- Required Error class in model.rb and null_adapter.rb
- Replaced all ::StandardError with Lotus::Model::Error
- Added tests

### DIFF
--- a/lib/lotus/model.rb
+++ b/lib/lotus/model.rb
@@ -4,6 +4,7 @@ require 'lotus/entity/dirty_tracking'
 require 'lotus/repository'
 require 'lotus/model/mapper'
 require 'lotus/model/configuration'
+require 'lotus/model/error'
 
 module Lotus
   # Model
@@ -16,7 +17,7 @@ module Lotus
     # @since 0.1.0
     #
     # @see Lotus::Repository.update
-    class NonPersistedEntityError < ::StandardError
+    class NonPersistedEntityError < Lotus::Model::Error
     end
 
     # Error for invalid mapper configuration
@@ -25,13 +26,13 @@ module Lotus
     # @since 0.2.0
     #
     # @see Lotus::Configuration#mapping
-    class InvalidMappingError < ::StandardError
+    class InvalidMappingError < Lotus::Model::Error
     end
 
     # Error for invalid raw command syntax
     #
     # @since 0.5.0
-    class InvalidCommandError < ::StandardError
+    class InvalidCommandError < Lotus::Model::Error
       def initialize(message = "Invalid command")
         super
       end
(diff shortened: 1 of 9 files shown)


lib/lotus/model.rb
```diff
@@ -16,7 +17,7 @@ module Model
     # @since 0.1.0
     #
     # @see Lotus::Repository.update
```
jodosha: @karimmtarek Instead of requiring a separated file with that exception, can you please define it here? Thank you!

jodosha:
@karimmtarek This looks good, but for extra safety, can you please add `test/error_test.rb`? Where we unit test that `Lotus::Model::Error` inherits from `::StandardError`. Then add one unit test for each custom exception that we have in the framework, and assert that they inherit from `Lotus::Model::Error`.

Apart from this improvement, the PR looks good. Thank you very much! :+1:

karimmtarek:
@jodosha I moved the `Error` class inside `model.rb` though I think now the `Error` class seems not to load properly in some `modules`, @pascalbetz helped my out and looked at the code and proposed couple of suggestions(please refer to the chat room for the complete transcript):
- Move `Error` class to it's own file and require it when needed.
- Some restructuring will be required if we'll keep the `Error` class inside `model.rb`

Other than this, I added couple of tests in the error_test.rb please let me know if this is what you had in mind.

lib/lotus/model.rb
```diff
@@ -25,13 +28,13 @@ class NonPersistedEntityError < ::StandardError
     # @since 0.2.0
     #
     # @see Lotus::Configuration#mapping
-    class InvalidMappingError < ::StandardError
+    class InvalidMappingError < Error
```
joneslee85: :clap: nice work
while you are on it, can we add test for every exception class?

karimmtarek: I added couple of test [there](https://github.com/karimmtarek/model/blob/introduce-lotus-model-error/test/error_test.rb), I just need to know if I'm testing the right thing, can you please look into it and advice?

test/error_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'test_helper'
+
+describe Lotus::Model::Error do
+  it 'inherits from ::StandardError' do
+    Lotus::Model::Error.superclass.must_be StandardError
+  end
+
+  it 'parent to all custom exception' do
```
joneslee85: I don't think this test adds more value because you have already tested it in `test/integration/repository_test.rb`

lib/lotus/model.rb
joneslee85: I left a comment there, btw more integration (if possible) test is welcome

karimmtarek:
@jodosha the tests I added are failing, is it because I need to require a file(s) I'm not sure, can you please look into it and advise, thanks.

karimmtarek:
@jodosha please review this when ever you have some time, thanks.

jodosha:
@karimmtarek This looks good. Thank you! :+1:

PR #268, opened 2015-12-16
Reduce object allocation for CollectionCoercer#_compile! method

davydovanton:
Check object allocation for all called code from test ([blog post](http://blog.davydovanton.com/2015/12/15/stackprof-with-minitest/))

### Before
```
==================================
  Mode: object(1)
  Samples: 1077870 (0.00% miss rate)
  GC: 0 (0.00%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
    445591  (41.3%)      238577  (22.1%)     Lotus::Model::Mapping::CollectionCoercer#_compile!
    104712   (9.7%)       78534   (7.3%)     Lotus::Utils::String#tokenize
```

### After
```
==================================
  Mode: object(1)
  Samples: 1007344 (0.00% miss rate)
  GC: 0 (0.00%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
    400814  (39.8%)      203930  (20.2%)     Lotus::Model::Mapping::CollectionCoercer#_compile!
     99240   (9.9%)       74430   (7.4%)     Lotus::Utils::String#tokenize
```

:tada:

### DIFF
--- a/lib/lotus/model/mapping/collection_coercer.rb
+++ b/lib/lotus/model/mapping/collection_coercer.rb
@@ -45,32 +45,41 @@ module Lotus
         # @api private
         # @since 0.1.0
         def _compile!
-          code = @collection.attributes.map do |_,attr|
-            %{
-            def deserialize_#{ attr.mapped }(value)
-              #{ attr.load_coercer }(value)
-            end
+          code = ''
+          hash = ''
+          taped_hash = ''
+          from_record = ''
+
+          @collection.attributes.each do |name, attr|
+            code << %{
+              def deserialize_#{attr.mapped}(value)
+                #{attr.load_coercer}(value)
+              end
             }
-          end.join("\n")
+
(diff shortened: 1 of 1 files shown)


jodosha:
@davydovanton Hey, thanks for this PR, but I'm not keen to merge it.

Before `lotusrb-0.6.0` Lotus::Model is "frozen", we don't wanna change anything. After that release, we want to change the way it works.

I don't like how `Coercer` is implemented, because of metaprogramming, and this PR touches that code.

This isn't a huge benefit in terms of speed/object allocations gain, because it change the "loading" code, not the "runtime" code. The former is ran only once at the application startup, even if saves object allocations, it doesn't affect the overall performance. The "runtime" code is the one that is executed for each incoming request. By improving it, we have a gain.

Sorry about this, I hope you will understand. I encourage you to contribute to Lotus with these perf patches. Thanks Anton!

davydovanton:
No problem, thanks for explanation!

PR #270, opened 2015-12-22
Raise error when try use migrator class if the adapter isn't a db adapter

AlfonsoUceda:
This PR resolves lotus/lotus#417 

Rescue `Sequel::AdapterNotFound` when we use a non database adapter and raise a `MigrationError` class displaying a message that the current adapter doesn't support database operations

cc @lotus/core-team

### DIFF
--- a/lib/lotus/model/migrator.rb
+++ b/lib/lotus/model/migrator.rb
@@ -292,6 +292,8 @@ module Lotus
         Sequel.connect(
           configuration.adapter.uri
         )
+      rescue Sequel::AdapterNotFound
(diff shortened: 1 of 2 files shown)


lib/lotus/model/migrator.rb
```diff
@@ -292,6 +292,8 @@ def self.connection
         Sequel.connect(
           configuration.adapter.uri
         )
+      rescue Sequel::AdapterNotFound
+        raise MigrationError.new("Current adapter (#{ configuration.adapter.type }) doesn't support database sequel operations.")
```
joneslee85: What do you think about rewording it to `does not support SQL database operations`?

AlfonsoUceda: :+1:

PR #281, opened 2016-01-15
Rescue and transform sequel database errors in Lotus::Command

Moratorius:
When creating an entity, the SqlAdapter does not raise a Lotus::Model::Error and
instead bubbles exceptions up from the adapter.  Applications should not be
concerned with the underlying adapter implementation, so the sql adapter command
has been modified to transform sequel database errors into lotus errors.

### DIFF
--- a/lib/lotus/model/adapters/sql/command.rb
+++ b/lib/lotus/model/adapters/sql/command.rb
@@ -30,7 +30,9 @@ module Lotus
           # @api private
           # @since 0.1.0
           def create(entity)
-            @collection.insert(entity)
+            _rescue_database_error do
+              @collection.insert(entity)
+            end
           end
 
           # Updates the corresponding record for the given entity.
@@ -42,7 +44,9 @@ module Lotus
           # @api private
           # @since 0.1.0
           def update(entity)
-            @collection.update(entity)
+            _rescue_database_error do
+              @collection.update(entity)
+            end
           end
 
           # Deletes all the records for the current query.
@@ -55,10 +59,41 @@ module Lotus
           # @api private
           # @since 0.1.0
           def delete
-            @collection.delete
+            _rescue_database_error do
+              @collection.delete
+            end
           end
 
           alias_method :clear, :delete
+
+          private
+
+          # Rescues Sequel::DatabaseError in yielded block
+          #
+          # @api private
+          def _rescue_database_error
+            yield
+          rescue Sequel::DatabaseError => e
+            raise _transform_database_error(e)
+          end
+
+          # Transforms error into a Lotus::Model::Error
+          #
+          # @api private
+          def _transform_database_error(error)
+            case error
+            when Sequel::CheckConstraintViolation
+              Lotus::Model::CheckConstraintViolationError.new(error.message)
+            when Sequel::ForeignKeyConstraintViolation
+              Lotus::Model::ForeignKeyConstraintViolationError.new(error.message)
+            when Sequel::NotNullConstraintViolation
+              Lotus::Model::NotNullConstraintViolationError.new(error.message)
+            when Sequel::UniqueConstraintViolation
+              Lotus::Model::UniqueConstraintViolationError.new(error.message)
+            else
+              Lotus::Model::InvalidCommandError.new(error.message)
+            end
+          end
         end
       end
     end
--- a/lib/lotus/model/adapters/sql/command.rb
+++ b/lib/lotus/model/adapters/sql/command.rb
@@ -30,9 +30,9 @@ module Lotus
           # @api private
           # @since 0.1.0
           def create(entity)
-            _rescue_database_error do
-              @collection.insert(entity)
-            end
+            @collection.insert(entity)
+          rescue Sequel::DatabaseError => e
+            raise Lotus::Model::Error.new(e.message)
           end
 
           # Updates the corresponding record for the given entity.
@@ -44,9 +44,9 @@ module Lotus
           # @api private
           # @since 0.1.0
           def update(entity)
-            _rescue_database_error do
-              @collection.update(entity)
-            end
+            @collection.update(entity)
+          rescue Sequel::DatabaseError => e
+            raise Lotus::Model::Error.new(e.message)
           end
 
           # Deletes all the records for the current query.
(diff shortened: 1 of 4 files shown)


lib/lotus/model/adapters/sql/command.rb
```diff
@@ -55,10 +59,41 @@ def update(entity)
...
           end
 
           alias_method :clear, :delete
+
+          private
+
+          # Rescues Sequel::DatabaseError in yielded block
+          #
+          # @api private
+          def _rescue_database_error
```
joneslee85: in theory private function should not be tested. But this function is an exception due to its richness in logic. I think you should remove the `private` syntax.

lib/lotus/model/adapters/sql/command.rb
```diff
@@ -55,10 +59,41 @@ def update(entity)
...
+            when Sequel::CheckConstraintViolation
+              Lotus::Model::CheckConstraintViolationError.new(error.message)
+            when Sequel::ForeignKeyConstraintViolation
+              Lotus::Model::ForeignKeyConstraintViolationError.new(error.message)
+            when Sequel::NotNullConstraintViolation
+              Lotus::Model::NotNullConstraintViolationError.new(error.message)
+            when Sequel::UniqueConstraintViolation
+              Lotus::Model::UniqueConstraintViolationError.new(error.message)
+            else
+              Lotus::Model::InvalidCommandError.new(error.message)
```
joneslee85: Can you think of any other type of error that Sequel could raise that does not fit into `InvalidCommandError`? What do you think about using `Lotus::Model::Error` here instead?

test/model/adapters/sql/command_test.rb
```diff
@@ -0,0 +1,106 @@
...
+    query.define_singleton_method(:scoped, -> { @collection })
+    query
+  end
+
+  before do
+    @command = Lotus::Model::Adapters::Sql::Command.new(query)
+  end
+
+  describe '#create' do
+    it 'rescues database errors' do
```
joneslee85: The spec description seems to be mismatching to the assertion. I think instead of testing the private method `rescue_database_error`, you should test it it with the public interface.

lib/lotus/model/adapters/sql/command.rb
```diff
@@ -55,10 +59,41 @@ def update(entity)
...
+          rescue Sequel::DatabaseError => e
+            raise _transform_database_error(e)
+          end
+
+          # Transforms error into a Lotus::Model::Error
+          #
+          # @api private
+          def _transform_database_error(error)
+            case error
+            when Sequel::CheckConstraintViolation
```
jodosha: @Moratorius I think all these cases can be handled by `Lotus::Model::Error`. So you can remove entirely this method.
```

File: lib/lotus/model/adapters/sql/command.rb [hunk 5]
```diff
@@ -55,10 +59,41 @@ def update(entity)
...
+
+          private
+
+          # Rescues Sequel::DatabaseError in yielded block
+          #
+          # @api private
+          def _rescue_database_error
+            yield
+          rescue Sequel::DatabaseError => e
+            raise _transform_database_error(e)
```
jodosha: @Moratorius I'd go for:

```ruby
raise Lotus::Model::Error.new(e.message)
```

That's it.

Starting from Ruby 2.1, exceptions respond to `#cause`, which returns the original raised exception. For our case, our `Lotus::Model::Error#cause` will return one of the `Sequel` raised exceptions. Read more at: http://devblog.avdi.org/2013/12/25/exception-causes-in-ruby-2-1/

lib/lotus/model/adapters/sql/command.rb
Moratorius: Now that the rescue has been simplified to raising one error, I'll just define a rescue for each of the public methods.

lib/lotus/model/adapters/sql/command.rb
Moratorius: I was preserving the behavior in `SqlAdapter#execute` (https://github.com/lotus/model/blob/v0.5.1/lib/lotus/model/adapters/sql_adapter.rb#L244-L246), where all `Sequel::DatabaseError` were transformed into `Lotus::InvalidCommandError`.  However, given https://github.com/lotus/model/pull/281#discussion_r49831938, I'll just raise `Lotus::Model::Error` irregardless of the specific error subclass.

lib/lotus/model/adapters/sql/command.rb
Moratorius: I agree, I did not know `#cause` existed.  Thanks!

davydovanton:
@Moratorius what do you think about this idea:

1. Create private method like this (name is terrible, I know :disappointed:):
``` ruby
def model_error_wrapper
  yield
rescue Error => e
  raise NewError
end
```
2. Rewrite `#create`, `#update`, `#delete`, etc using `model_error_wrapper`:
``` ruby
def create(entity)
  model_error_wrapper { @collection.insert(entity) }
end
```

/cc @jodosha, @joneslee85

lib/lotus/model/adapters/sql/command.rb
Moratorius: However, not providing an abstracted error means that applications will have to be coupled directly to the adapter-specific error.  For example, if the application wanted to do something when an entity is not persisted because of a uniqueness constraint, the application would have to inspect the `Lotus::Model::Error#cause` for `Sequel::UniqueConstraintViolationError`.  This seems to defeat the purpose of the adapter abstraction.  What do you suggest applications do in such cases?

joneslee85: @Moratorius i am too not a big fan of this pattern. I prefer more specific exception for each case like you did originally. But the decision in the end is up to @jodosha.

joneslee85:
@Moratorius i am too not a big fan of this pattern. I prefer more specific exception for each case like you did originally. But the decision in the end is up to @jodosha.

@davydovanton the code can ben DRY a bit more but firstly it is needed to be decided which path it should take

test/model/adapters/sql/command_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    @command = Lotus::Model::Adapters::Sql::Command.new(query)
+  end
+
+  describe '#create' do
+    describe 'when a Sequel::DatabaseError is raised' do
+      it 'raises a lotus error' do
+        collection.define_singleton_method(:insert) do |_|
+          raise Sequel::DatabaseError.new
+        end
+        -> { @command.create(Object.new) }.must_raise(Lotus::Model::Error)
```
jodosha: @Moratorius Hey, this looks good :+1: Extra bonus points if we assert the message:

```ruby
exception = -> { @command.create(Object.new) }.must_raise(Lotus::Model::Error)
exception.message.must_equal '...'
```

What do you think?

lib/lotus/model/adapters/sql/command.rb
cyphactor: @jodosha would you please explain the stance you are taking on this and the reasoning behind it. I would love to understand your thoughts and reasoning both from the pragmatic side of things as well as the purest design side of things so that I and others can make these decisions prior to PR in the future.

Issue #280, opened 2016-01-13
Show message about including `.load!` after config


cllns:
I got this following error, even though I had set both `adapter` and `mapping`
```
#<Lotus::Model::Adapters::NoAdapterError: Cannot invoke `create' on repository. Please check if `adapter' and `mapping' are set.>
```

The problem was that the end of the config block didn't call `.load!`. 

It'd be nice if the error handling checked to see if that was the problem, and provided a specific error message in that case!

AlfonsoUceda:
@cllns thanks for the issue, could you make a PR?

jodosha:
@cllns Okay, In our docs we specify a lot of times that `.load!` should be invoked before to use it. See [README](https://github.com/lotus/model/blob/master/README.md). In `lotusrb`, the generated code does it for you.

We're open for an improved error message, unless it requires to change too much the internal code. Thank you very much :smile:

cllns:
Closed via #282 !

PR #282, opened 2016-01-18
Improve NoAdapterError by mentioning `.load!`

cllns:
Addresses #280.

I wrote code to solve this issue (by adding a `.loaded?` method to Lotus::Model::Configuration), but it added too much code, based on @jodosha's direction that he doesn't want to add a lot of code for this minor improvement.

I think this is better anyway, since it makes the error message more complete. Configurations need **3** things, and now the error message references all of them:

1. `adapter`
2. `mapping`
3. `.load!`

Also, I prefer multi-line strings to long lines, but I can switch them to long single lines if that's preferable.

### DIFF
--- a/lib/lotus/model/adapters/null_adapter.rb
+++ b/lib/lotus/model/adapters/null_adapter.rb
@@ -6,7 +6,9 @@ module Lotus
       # @since 0.2.0
       class NoAdapterError < Lotus::Model::Error
         def initialize(method_name)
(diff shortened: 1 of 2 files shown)


cllns:
Looks like the Travis failure is unrelated to my change :)

joneslee85:
@cllns looks legit, I'll restart the build for you

PR #283, opened 2016-01-19
Add test to assert exception message

joneslee85:
Related to #281

### DIFF
--- a/test/model/adapters/sql/command_test.rb
+++ b/test/model/adapters/sql/command_test.rb
@@ -14,61 +14,107 @@ describe Lotus::Model::Adapters::Sql::Command do
   end
 
   describe '#create' do
-    describe 'when a Sequel::DatabaseError is raised' do
-      it 'raises a lotus error' do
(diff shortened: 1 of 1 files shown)


jodosha:
@joneslee85 In the original implementation of #281, @Moratorius used concrete examples of exceptions that Sequel can raise. They carry on useful messages for developers.

There is no value for them to see `"Sequel::Error"` without having a clue on what to do.

This PR should solve this problem. Can you please try to raise realistic exceptions and check if the returning message is the same? Thank you. :+1: 

See https://github.com/Acornsgrow/model/commit/1db52737f0a3c00964f73ba7e94e1f7433b9c6f8#diff-1ad76a79f04e29c81971a6206b63b136R85

jodosha:
@joneslee85 But we aren't testing against `Sequel::CheckConstraintViolation` and others exceptions. That was the point of this cleanup :wink:

jodosha:
@joneslee85 Sweet, thank you very much :+1:

PR #289, opened 2016-02-01
Mapping SQL Adapter's Errors into specific Model's Error

### LINKED ISSUE #284: Other types of Lotus::Model::Error for some specific adapter exceptions
Trying out Lotus Model v0.5.2 here, I've found that Lotus::Model::Error missed a more specialized way to deal with Sequel Exceptions.

For example: I need to deal with `Sequel::UniqueConstraintViolation` for any repository I might have. My PK’s are Postgres’ _uuid fields_ which throws that exception if I try to query any data but UUID Format. Something like this:

```ruby
my_non_uuid = 1234
MyRepository.find(my_non_uuid) # fail with Sequel::UniqueCo…
MyRepository.find('82994efa-5d21-4e8a-9a25-0c9448b91f4e') # nil or Object :+1:
```

In 0.5.2, I will get a generic `Lotus::Model::Error` which make a bit harder to identify this kind of contract violation of my requester. I am not suggesting to Lotus deals with _uniqueness constraints_. Just a (IMO) _better_ encapsulation over application layer/tiers.

TL;DR: my suggestion downs to: more (and specialized) Lotus Model exceptions.

hlegius:
As discussed in #284, we agreed to rollback the adapter violation errors in prior to (re)raise specific errors (i.e. Unique Constraints, ForeignKey Violations et al) encapsulated by our Model Errors.

I've created the new error classes inside `hanami/model.rb`, but I'd rather place them (and move all of them) inside `hanami/model/error.rb` just like `Hanami::Model::Error`. /cc @hanami/core-team 

Related #281 
Closes #284

### DIFF
--- a/lib/hanami/model/adapters/sql/command.rb
+++ b/lib/hanami/model/adapters/sql/command.rb
@@ -30,9 +30,7 @@ module Hanami
           # @api private
           # @since 0.1.0
           def create(entity)
-            @collection.insert(entity)
-          rescue Sequel::DatabaseError => e
-            raise Hanami::Model::Error.new(e.message)
+            _handle_database_error { @collection.insert(entity) }
           end
 
           # Updates the corresponding record for the given entity.
@@ -44,9 +42,7 @@ module Hanami
           # @api private
           # @since 0.1.0
           def update(entity)
-            @collection.update(entity)
-          rescue Sequel::DatabaseError => e
-            raise Hanami::Model::Error.new(e.message)
+            _handle_database_error { @collection.update(entity) }
           end
 
           # Deletes all the records for the current query.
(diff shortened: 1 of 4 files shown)


lib/hanami/model/adapters/sql/command.rb
```diff
@@ -59,15 +55,45 @@ def update(entity)
...
+              Hanami::Model::CheckConstraintViolationError.new(error.message)
+            when Sequel::ForeignKeyConstraintViolation
+              Hanami::Model::ForeignKeyConstraintViolationError.new(error.message)
+            when Sequel::NotNullConstraintViolation
+              Hanami::Model::NotNullConstraintViolationError.new(error.message)
+            when Sequel::UniqueConstraintViolation
+              Hanami::Model::UniqueConstraintViolationError.new(error.message)
+            else
+              Hanami::Model::InvalidCommandError.new(error.message)
+            end
```
jodosha: @hlegius What do you think about this syntax?

```ruby
case error
when Sequel::CheckConstraintViolation
  Hanami::Model::CheckConstraintViolationError
when Sequel::ForeignKeyConstraintViolation
  Hanami::Model::ForeignKeyConstraintViolationError
when Sequel::NotNullConstraintViolation
  Hanami::Model::NotNullConstraintViolationError
when Sequel::UniqueConstraintViolation
  Hanami::Model::UniqueConstraintViolationError
else
  Hanami::Model::InvalidCommandError
end.new(error.message)
```

hlegius: Good! :ok_hand:

jodosha:
@hlegius This looks good :+1:

joneslee85:
:+1:

lib/hanami/model/adapters/sql/command.rb
joneslee85: On second look, what do you think about storing them into a frozen hash the query them by the key, should the pair does not match the specified key, the code fallback to the default class.

```
MAPPING = {...}.freeze

(MAPPING[error] || InvalidCommandError).new(error.message)
```

davydovanton: Nice solution @joneslee85. I like it :+1:

hlegius: @joneslee85 Good idea! I will push another PR with this modification plus some other related stuff pending to do around this.

PR #299, opened 2016-02-03
Moving class level methods to instance methods for Repositories

pascalbetz:
See https://github.com/hanami/model/issues/290

* Keep class level accessors for _adapter_ and _collection_
* Add instance level accessors for _adapter_ and _collection_ that use
  class level accessors
* Move class to instance level methods
* Change all tests and sample code to reflect changes

### DIFF
--- a/lib/hanami/repository.rb
+++ b/lib/hanami/repository.rb
@@ -130,10 +130,11 @@ module Hanami
     #     include Hanami::Repository
     #   end
     def self.included(base)
+
       base.class_eval do
+        include InstanceMethods
         extend ClassMethods
         include Hanami::Utils::ClassAttribute
-
         class_attribute :collection
         self.adapter = Hanami::Model::Adapters::NullAdapter.new
       end
@@ -208,14 +209,27 @@ module Hanami
       #   end
       #
       #   UserRepository.adapter = adapter
+      # @since 0.5.0
+      # @api private
       def adapter=(adapter)
         @adapter = adapter
       end
 
-      # @since 0.5.0
-      # @api private
       def adapter
         @adapter
+      end
+
+    end
+
+    module InstanceMethods
+      # Get the currently configured adapter
+      def adapter
+        self.class.adapter
+      end
+
+      # Get the currently configured collection name
+      def collection
+        self.class.collection
       end
 
       # Creates or updates a record in the database for the given entity.
@@ -239,7 +253,7 @@ module Hanami
       #   article = Article.new(title: 'Introducing Hanami::Model')
       #   article.id # => nil
       #
-      #   persisted_article = ArticleRepository.persist(article) # creates a record
+      #   persisted_article = ArticleRepository.new.persist(article) # creates a record
       #   article.id # => nil
       #   persisted_article.id # => 23
       #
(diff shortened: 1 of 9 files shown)


davydovanton:
good job :+1:

hlegius:
:ok_hand:

joneslee85:
:+1:

joneslee85:
@pascalbetz please rebase

pascalbetz:
@joneslee85 rebase done.

lib/hanami/model.rb
```diff
@@ -11,6 +11,78 @@ module Hanami
...
   # @since 0.1.0
   module Model
+    # Error for non persisted entity
+    # It's raised when we try to update or delete a non persisted entity.
+    #
+    # @since 0.1.0
+    #
+    # @see Hanami::Repository.new.update
+    class NonPersistedEntityError < Hanami::Model::Error
+    end
```
joneslee85: I think this is not related to this PR. Seems a bad rebase?

hlegius: Yep. It was moved to `model/error.rb`. You can safely remove these lines.

lib/hanami/repository.rb
```diff
@@ -208,16 +209,33 @@ module ClassMethods
       #   end
       #
       #   UserRepository.adapter = adapter
+      # @since 0.5.0
+      # @api private
       def adapter=(adapter)
         @adapter = adapter
       end
 
-      # @since 0.5.0
```
joneslee85: Shouldn't the comments removed?

lib/hanami/repository.rb
```diff
@@ -208,16 +209,33 @@ module ClassMethods
...
       def adapter
         @adapter
       end
 
+    end
+
+    module InstanceMethods
+      # Get the currently configured adapter
+      #
+      # Returns the adapter that is specified on the class level
```
joneslee85: I do not think we need this comment line since the code reveals the intent well enough

lib/hanami/repository.rb
pascalbetz: Good question. Actually I was thinking about removing/resetting all the ```@version``` comments as well for the ```InstanceMethods``` methods that were previous ```ClassMethods```

About removing the comments: it's still valid. Just not the preferred way anymore. But until we have decided what the new preferred way is, i'd leave the comments.

test/integration/repository_test.rb
```diff
@@ -589,7 +611,7 @@ def updated_at
       end
 
       it "raises an exception because memory adapter doesn't support fetch" do
-        -> { ArticleRepository.find_raw }.must_raise NotImplementedError
+        -> { ArticleRepository.new.find_raw }.must_raise NotImplementedError
```
joneslee85: can we asset the exception message too?

joneslee85:
except comments above, this is a good PR

lib/hanami/repository.rb
joneslee85: I see. Fair point

joneslee85:
cc @AlfonsoUceda

lib/hanami/repository.rb
```diff
@@ -208,16 +209,29 @@ module ClassMethods
...
-      # @since 0.5.0
-      # @api private
       def adapter
         @adapter
       end
 
+    end
+
+    module InstanceMethods
+      # Get the currently configured adapter
```
AlfonsoUceda: could you add `@since x.x.x` and below?

AlfonsoUceda:
:+1: we will see where we end with this changes :D

lib/hanami/repository.rb
jodosha: @pascalbetz @joneslee85 Yes, class => instance methods should be updated with `@version x.x.x`.

jodosha:
@pascalbetz This looks good to me. Merging :+1:

PR #300, opened 2016-02-04
Use fetch instead of [] and || for default value.

pascalbetz:
Feedback from Codereview.

* Use ```fetch``` instead of ```[]``` and ```||``` for default value
* More descriptive name


davydovanton:
nice solution :+1:

joneslee85:
nice code

PR #301, opened 2016-02-14
Fixed: Repository#find for wrong types will not raise error

### LINKED ISSUE #287: Trying to find using unexpected type
I have a JSON API with primary keys of all entities of Postgres UUID type. Users search a lot by ids, and the problem happens when user occasionally enter random string instead of proper UUID.

```ruby
> BooksRepository.find('invalid')
Hanami::Model::InvalidQueryError: PG::InvalidTextRepresentation: ERROR:  invalid input syntax for uuid: "invalid"
LINE 1: SELECT * FROM "books" WHERE ("id" = 'invalid...
```

I'd rather expect this request returns just `nil`, instead of type error. I think controlling such type of things is pretty verbose and annoying.

hlegius:
Before:
```ruby
BookRepository.new.find(1)   # => book
BookRepository.new.find('1') # => book

BookRepository.new.find(999) # => nil
BookRepository.new.find('999') # => nil

# For Integer PKs
BookRepository.new.find('invalid')
# => TypeError: can't convert "invalid" into Integer

# For UUID Primary Keys (SQL)
BookRepository.new.find(1123)
# => Hanami::Model::InvalidQueryError: PG::InvalidTextRepresentation: ERROR:  invalid input syntax for uuid: "invalid"
# => LINE 1: SELECT * FROM "books" WHERE ("id" = '1123...
```

Now:

```ruby
BookRepository.new.find(1)   # => Book
BookRepository.new.find('1') # => Book

BookRepository.new.find(999) # => nil
BookRepository.new.find('999') # => nil

# For Integer PKs
BookRepository.new.find('invalid') # => nil

# For UUID Primary Keys (SQL)
BookRepository.new.find(1123) # => nil
```

Closes #287

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,12 @@
 language: ruby
 sudo: false
 cache: bundler
-install: true
-script:
-  - 'if [[ "$TRAVIS_RUBY_VERSION" =~ "jruby" ]]; then rvm get head && rvm reload && rvm use --install $TRAVIS_RUBY_VERSION; fi'
-  - 'bundle install'
-  - 'bundle exec rake test:coverage'
+script: 'bundle exec rake test:coverage --trace'
 rvm:
   - 2.2.4
   - 2.3.0
+  - jruby-9.0.5.0
   - ruby-head
-  - jruby-9000
   - jruby-head
 
 addons:
--- a/lib/hanami/model/adapters/implementation.rb
+++ b/lib/hanami/model/adapters/implementation.rb
@@ -50,6 +50,8 @@ module Hanami
           _first(
             _find(collection, id)
           )
+        rescue TypeError, Hanami::Model::InvalidQueryError
+          nil
         end
 
         # Returns the first record in the given collection.
--- a/lib/hanami/model/adapters/implementation.rb
+++ b/lib/hanami/model/adapters/implementation.rb
@@ -93,7 +93,7 @@ module Hanami
         end
 
         def _type_mismatch_error?(error)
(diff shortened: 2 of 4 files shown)


jodosha:
@hlegius Thank for this fix. While this PR solves the target issue, I'm a bit afraid that suppressing high level exceptions such as `TypeError` and `InvalidQuery` can lead to unexpected behaviors.

Do we have any hint form the low level stack and suppress only when we're sure that the exception is caused by this specific problem?

hlegius:
@jodosha I see and felt the same thing while studying it. Actually, I've started from there (inner levels), but I've found some _problems_: `find` calls `_find`which calls `_id`. All these surrounded by `_first` which expect a `query` to perform. The first two of them are specific for the public interface (`find`) and both if return 'nil' can lead towards unexpected behaviour like unexpected message error, because both of them combine part of the adapter's query implementation.

I went for suppress both errors on `find` because it has a specific input an defined query as output. I did some additional tests (manually within my test project) and got no _weird_ behaviour.

If you think its more secure, I can parse both exceptions to return nil only if is about casting error. What do you think?

Thanks for the review :100: 

btw, we're with some JRuby issue on CI :/

jodosha:
> If you think its more secure, I can parse both exceptions to return nil only if is about casting error. What do you think?

That could be better.

> btw, we're with some JRuby issue on CI :/

Can you please bump to `9.0.5.0`? Please have a look at `utils` `.travis.yml`.

jodosha:
@hlegius This looks good. Merging :+1:

Issue #291, opened 2016-02-02
Make repositories to work with data


jodosha:
## Actual

Right now we only accept entities as input for repositories default CRUD operations. This often leads to awkward and unnecessary objects allocations.

``` ruby
book = BookRepository.create(
  Book.new(params[:book])
)
```

That intermediate line is unnecessary. We should adopt the _"data in, entities out"_ principle.
## Proposal

Default CRUD operations should accept any object that respond to `#to_hash`.

``` ruby
book = BookRepository.create(params[:book])
```

To maintain backward compatibility we should implement `Hanami::Entity#to_hash`. In this way, the syntax above will still work. Please notice that `Hanami::Entity` already has `#to_h`, so it's just a matter of alias that method.

All the other objects will work just fine:

``` ruby
➜ pry
[1] pry(main)> require 'rubygems'
=> false
[2] pry(main)> require 'hanami/controller'
=> true
[3] pry(main)> Hanami::Action::Params.new({}).respond_to?(:to_hash)
=> true
[4] pry(main)> {}.respond_to?(:to_hash)
=> true
```

---

The decision to use `#to_hash` over `#to_h` is demanded by the dangerous implementation that some objects like `Array` have.

``` ruby
➜ pry
[1] pry(main)> [].to_h
=> {}
[2] pry(main)> [1].to_h
TypeError: wrong element type Fixnum at 0 (expected array)
from (pry):2:in `to_h'
[3] pry(main)> [['a', 1]].to_h
=> {"a"=>1}
```

There is the risk that a `TypeError` could be raised if an array is passed. Better to rely on `#to_hash`, and keep `Array` out because:

``` ruby
➜ pry
[1] pry(main)> [].respond_to?(:to_hash)
=> false
```

booch:
I hadn't considered the "data in, entities out" idea, but it looks like a good idea. Given that we can still use an entity as input, I don't see the harm in the proposed change.

joneslee85:
Seems a safe change to me.

What's data?

What type of hash should  `to_hash` return? A string vs symbol key hash?

Bounga:
It look really good to me :+1:

It's a huge step forward in term of flexibility.

@joneslee85 I'd like to be able to use both. I think it's simple to handle.

jodosha:
@joneslee85 

> What type of hash should to_hash return? A string vs symbol key hash?

This is an excellent question that opens a new, huge topic. Please keep in mind that most of the times data means a Hash that is coming from a Rack env.

[...]

PR #302, opened 2016-02-16
Make repositories work with data in, entities out

AlfonsoUceda:
This PR implements issue #291 

### TODO
- [ ] Update doc

### DIFF
--- a/lib/hanami/entity.rb
+++ b/lib/hanami/entity.rb
@@ -219,11 +219,9 @@ module Hanami
     #
     #   user = User.new(id: 23, name: 'Luca')
     #   user.to_h # => { :id => 23, :name => "Luca" }
-    #   user.to_hash # => { :id => 23, :name => "Luca" }
     def to_h
       Hash[attribute_names.map { |a| [a, read_attribute(a)] }]
     end
-    alias :to_hash :to_h
 
     # Return the set of attribute names
     #
--- a/lib/hanami/entity.rb
+++ b/lib/hanami/entity.rb
@@ -219,9 +219,11 @@ module Hanami
     #
     #   user = User.new(id: 23, name: 'Luca')
     #   user.to_h # => { :id => 23, :name => "Luca" }
+    #   user.to_hash # => { :id => 23, :name => "Luca" }
     def to_h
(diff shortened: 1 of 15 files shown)


lib/hanami/entity.rb
```diff
@@ -219,9 +219,11 @@ def ==(other)
     #
     #   user = User.new(id: 23, name: 'Luca')
     #   user.to_h # => { :id => 23, :name => "Luca" }
+    #   user.to_hash # => { :id => 23, :name => "Luca" }
     def to_h
       Hash[attribute_names.map { |a| [a, read_attribute(a)] }]
```
joneslee85: Should we use `Hanami::Utils::Hash`?

test/integration/repository_test.rb
```diff
@@ -107,6 +107,18 @@
...
           end
+
+          it 'works with entity' do
+            article = ArticleRepository.new.persist(Article.new(title: 'Art 1'))
+            article.must_be_kind_of Article
+            article.id.wont_be_nil
+          end
+
+          it 'works with hash' do
+            article = ArticleRepository.new.persist({ title: 'Art 1' })
```
joneslee85: worth to have a test for string key hash

lib/hanami/entity.rb
jodosha: @joneslee85 This won't make `#to_h` compatible with the rest of Ruby ecosystem, which doesn't know about our `Utils::Hash`.

test/integration/repository_test.rb
jodosha: Yes both :+1:

AlfonsoUceda:
I had problems with rebase, I'll create another PR when this is done.

lib/hanami/entity.rb
```diff
@@ -219,9 +219,11 @@ def ==(other)
...
     end
+    alias :to_hash :to_h
```
AlfonsoUceda: @jodosha I think we can remove this changes now, what do you think?

jodosha: @AlfonsoUceda Yes. :+1:

jodosha:
@AlfonsoUceda Closing this in favor of a new implementation I'm working on. Thanks!

PR #303, opened 2016-02-20
Fix SimpleCov deprecation message

deepj:
This avoids the following deprecation message:

```
test/test_helper.rb:8:in `<top (required)>': [DEPRECATION] ::[] is deprecated. Use ::new instead.
```


deepj:
Ugh, I'm now sure what's wrong with coverage. Decreased by -0.0% and it failed?

PR #305, opened 2016-02-29
move Model.migration to model.rb

jbodah:
Personal preference, but I think this would be easier to lookup the definition of `Hanami::Model::migration` if it was defined in `lib/hanami/model.rb` instead of re-opening `Hanami::Model` in `lib/hanami/model/migrator.rb`. I'm not sure how this would compare to the rest of the code however

Thoughts?

I'm currently having trouble reproducing the test environment and don't have time at the moment to debug anything, so I'm relying on Travis for testing

### DIFF
--- a/lib/hanami/model.rb
+++ b/lib/hanami/model.rb
@@ -168,5 +168,46 @@ module Hanami
         duplicated.configure(&blk) if block_given?
       end
     end
+
+    # Define a migration
+    #
+    # It must define an up/down strategy to write schema changes (up) and to
+    # rollback them (down).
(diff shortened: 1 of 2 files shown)


jodosha:
@jbodah Hey thanks for this PR. There is an intention to keep it outside of `model.rb`. Migrations only work for SQL databases. We're looking for ways to transform `sequel` from runtime dependency, to soft dependency to be installed only when you work with a SQL database.

The actual file separation, already gives the structure for this change. I'm closing this PR. Thanks!

PR #307, opened 2016-03-24
Improve error-handling for adapter

cllns:
1. Reclassify Adapter errors as Hanami::Model::Error
2. Leverage the `AdapterNotFound` Error class, instead of using `LoadError`

https://github.com/hanami/hanami/issues/556 item 1

### DIFF
--- a/lib/hanami/model/config/adapter.rb
+++ b/lib/hanami/model/config/adapter.rb
@@ -7,8 +7,8 @@ module Hanami
       #
       # @since 0.2.0
       class AdapterNotFound < Hanami::Model::Error
-        def initialize(adapter_name)
-          super "Cannot find Hanami::Model adapter #{adapter_name}"
+        def initialize(adapter_name, message)
+          super "Cannot find Hanami::Model adapter '#{adapter_name}' (#{message})"
         end
       end
 
@@ -95,7 +95,7 @@ module Hanami
           begin
             require "hanami/model/adapters/#{type}_adapter"
           rescue LoadError => e
-            raise LoadError.new("Cannot find Hanami::Model adapter '#{type}' (#{e.message})")
+            raise AdapterNotFound.new(class_name, e.message)
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/hanami/model/config/adapter.rb
```diff
@@ -95,7 +95,7 @@ def load_adapter
           begin
             require "hanami/model/adapters/#{type}_adapter"
           rescue LoadError => e
-            raise LoadError.new("Cannot find Hanami::Model adapter '#{type}' (#{e.message})")
+            raise AdapterNotFound.new(class_name, e.message)
```
cllns: Leverages Hanami::Model::Error:AdapterNotFound custom error, instead of using LoadError

lib/hanami/model/config/adapter.rb
```diff
@@ -103,13 +103,10 @@ def instantiate_adapter(mapper)
           begin
             klass = Hanami::Utils::Class.load!(class_name, Hanami::Model::Adapters)
             klass.new(mapper, uri, options)
-          rescue NameError
-            raise AdapterNotFound.new(class_name)
```
cllns: Got rid of this because:
1. It hides NoMethodErrors, which are quite common because of `nil`...
2. The obvious purpose, to catch undefined classes, has a check in `load_adapter`, which is called right before this. They're both private methods, so the class should always be defined by the time the code gets here.

lib/hanami/model/config/adapter.rb
```diff
@@ -103,13 +103,10 @@ def instantiate_adapter(mapper)
...
           rescue => e
-            raise "Cannot instantiate adapter of #{klass} (#{e.message})"
+            raise Error, e
```
cllns: This preserves only the message from the original exception, not the location.

Instead we bubble the error up, casting it as a `Hanami::Model::Error`

jodosha:
@cllns Merged. Thank you!

PR #308, opened 2016-03-28
Raise MissingURIError for FileSystemAdapter is URI is nil or empty

cllns:
Addresses #3 for https://github.com/hanami/hanami/issues/556.

That issue was closed but we still need a helpful error for when URI is nil or empty. For example, if someone accidentally deletes their `DATABASE_URL` environment variable, we should help them realize that.

The current error doesn't help you identify that at all, even after #307: `/Users/sean/dev/hanami-model/lib/hanami/model/config/adapter.rb:107:in `rescue in instantiate_adapter': undefined method `sub' for nil:NilClass (Hanami::Model::Error)`

And now the error is: `/Users/sean/dev/hanami-model/lib/hanami/model/config/adapter.rb:107:in 'rescue in instantiate_adapter': URI for FileSystemAdapter is nil or empty (Hanami::Model::Error)`, which is more helpful.

Happy to move this to `AbstractAdapter`, since it's relevant for all of the adapters (except MemoryAdapter), although the SQL adapters will have their own errors for missing URI's.

### DIFF
--- a/lib/hanami/model/adapters/abstract.rb
+++ b/lib/hanami/model/adapters/abstract.rb
@@ -25,6 +25,15 @@ module Hanami
       class NotSupportedError < Hanami::Model::Error
       end
 
+      # It's raised when a URI is nil or empty
+      #
+      # @since x.x.x
+      class MissingURIError < Hanami::Model::Error
+        def initialize
+          super "URI for FileSystemAdapter is nil or empty"
+        end
+      end
+
       # It's raised when an operation is requested to an adapter after it was
       # disconnected.
       #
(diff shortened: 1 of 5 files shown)


jodosha:
@cllns Sorry for the late reply after the Easter break. :hatching_chick:

This improvement is just great. We need more and more take control of this kind of exceptions and return helpful errors for developers. :100: 

As you suggested, please make that logic available for all the adapters, via `AbstractAdapter`, then we can merge it. :+1:

cllns:
All set!

Another option is changing `uri.nil? || uri.empty?` to `!uri.respond_to(:to_s) || uri.empty?`

jodosha:
@cllns Thank you, I merged it :+1:

PR #311, opened 2016-04-05
Require 'json' before checking if to_json is allowed

cllns:
This test fails for me locally.

This test passes on Travis because it runs `bundle exec rake test:coverage` which sets `COVERAGE=true`. And `test_helper.rb` requires 'simplecov' and 'coveralls' when `COVERAGE` is true and they, in-turn, require 'json'.

This PR just requires 'json' right before checking to see if `to_json` is allowed.

### DIFF
--- a/test/entity_test.rb
+++ b/test/entity_test.rb
@@ -62,6 +62,7 @@ describe Hanami::Entity do
     end
 
     it 'returns false for reserved word' do
+      require "json"
       Car.allowed_attribute_name?(:to_json).must_equal false
     end
   end


jodosha:
@cllns Hello and thanks for this PR. Which OS are you using?

cllns:
This could be an issue on my end, just not sure what it'd be! It happens on Ruby 2.3.0, 2.2.2 and 2.1.6 for me.

Using OS X 10.11.2 and Postgres 9.4.7.0 (no MySQL)

@jodosha Do you have `COVERAGE=true` in your environment? That causes this to pass.

Does: 
```
COVERAGE=false bundle exec ruby -Ilib:test test/entity_test.rb
```
 show this failure?

cc/ @hanami/contributors if anyone wants to help out and run that command too :)

davydovanton:
mac os 10.11.2
I have a broken tests ([link](https://gist.github.com/davydovanton/8c89fb7fbc80dc85c668c30f4fd0bb00)) too.

jodosha:
@cllns Merged, thank you!

PR #312, opened 2016-04-07
Allow running specific adapter's tests, via ENV['ADAPTER']

### LINKED ISSUE #257: Improve tests

cllns:
Closes #257.

You can try this out by calling `ADAPTER=postgres bundle exec rake`.

One potential bug/feature of this is that running with `ADAPTER=dog` will silently run none of the adapter-specific tests.

It looks like I change `mysql_test.rb` and `postgres_test.rb` a lot but I just add a conditional around the whole file and indent inside of that.

### DIFF
--- a/test/integration/migration/mysql_test.rb
+++ b/test/integration/migration/mysql_test.rb
@@ -1,322 +1,324 @@
 require 'test_helper'
 require 'hanami/model/migrator'
 
-describe 'Mysql database migrations' do
-  let(:adapter_prefix) { 'jdbc:' if Hanami::Utils.jruby?  }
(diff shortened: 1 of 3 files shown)


jodosha:
@cllns Thanks for this PR, but we're probably gonna to extract SQL code from `hanami-model`. The proposal is useful if we would keep that code inside this gem, but this isn't the case.

Before to make this big change it testing code, I'd hold for now. Thanks for understanding.

PR #315, opened 2016-05-25
use map instead of inject

max-power:


jodosha:
@max-power Hey, thanks for this PR, but there are [upcoming changes](https://github.com/hanami/model/tree/new-engine) that need `Entity` to be reworked in its implementation.

PR #316, opened 2016-05-30
Fix problem with inheritance and Entitiy::attributes

kjellm:
Hi,

First of all, thank you all for the great work done on the Hanami project! Love using it :-)

There is a bug in the way the inheritance of attributes are implemented with regards to the `::attributes` method. The following example illustrates the problem:

```
class A
  include Hanami::Entity
  attributes :a
end

class B < A
  attributes :b
end

B.attributes # => [:id, :b], but should be [:id, :a, :b]
```

The current code adds the super class attributes to entities by calling ::define_attribute_accessor, but this fails silently because of  a guard on `!instance_methods.include?`

This pull request fixes the problem by adding super class attributes to the result when calling `::attributes`.

### DIFF
--- a/lib/hanami/entity.rb
+++ b/lib/hanami/entity.rb
@@ -140,14 +140,9 @@ module Hanami
       #   DeletedUser.attributes => #<Set: {:id, :name, :deleted_at}>
       #
       def attributes(*attrs)
-        return @attributes ||= Set.new unless attrs.any?
-
-        Hanami::Utils::Kernel.Array(attrs).each do |attr|
-          if allowed_attribute_name?(attr)
-            define_attr_accessor(attr)
-            self.attributes << attr
-          end
-        end
+        @attributes ||= Set.new
+        return get_attributes unless attrs.any?
+        set_attributes attrs
       end
 
       # Define setter/getter methods for attributes.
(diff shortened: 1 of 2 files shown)


jodosha:
@kjellm Thanks for this PR, but we're doing some internal rework for `hanami-model`: https://github.com/hanami/model/tree/new-engine

That will change how entities work, so I'm closing this PR. Thank you again 😄

kjellm:
@jodosha Would it be possible for you to apply the patch on the v0.6.1 tag, and then make a bug-fix release? I can make a new pull-request if it makes it easier for you.

jodosha:
@kjellm okay, deal. Here's the release: https://rubygems.org/gems/hanami-model/versions/0.6.2 😉

kjellm:
Awesome! Thanks :-)

kjellm:
@jodosha: I noticed that the release was yanked from rubygems.org. What happened? Anything I can do to help?

jodosha:
@kjellm Sorry for the confusion here. I had to yank both `0.6.2` and `0.6.2.1` from Rubygems.org because the `0.6.x.` branch contains breaking changes from master.

The first release was requiring `hanami-utils ~> 0.8`, which is unreleased yet. With the second one, I discovered the problem of the breaking changes. For instance https://github.com/hanami/model/pull/299 was released by mistake.

I'm afraid that `0.6.x` series is frozen.

Issue #322, opened 2016-06-17
Does Postgres Adapter display the correct error when db exists?


wjdhamilton:
Hello everyone!

When running `hanami db create` and there is already a database in existence, the error returned is as follows: 

```
createdb: database creation failed. There is 1 other session using the database.
```

A quick look at the code in model/lib/hanami/model/migrator/postgres_adapter.rb reveals the following snippet: 

``` ruby

message = if error_message.match(/already exists/)
              "createdb: database creation failed. There is 1 other session using the database."
```

Which implies to me that the error message is the same regardless of whether or not there really _is_ another 1 (or many more) sessions using the database. 

I would have expected that the app would differentiate between contexts when another connection is present and when it is simply the case that the db has already been created. Shouldn't the error message simply say that the db already exists if there are no sessions using the db?

cllns:
Thanks for reporting this!
This is a really good point. We definitely want accurate and helpful error messages.

Right now, @jodosha is working on re-working the back-end of `hanami-model` to be based on [Ruby Object Mapper (`rom`)](https://github.com/rom-rb/rom) and I *think* this code will be going away. 
So, I don't think it's worth creating a PR, but @jodosha can correct me if I'm wrong :)

wjdhamilton:
No problem! Glad to have been of service. I'll await @jodosha 😄

jodosha:
@wjdhamilton I think that this is fine to be fixed in master first **with the current code**. So the `new-engine` branch will already have working code.

@wjdhamilton or @cllns Can you please open a PR that fix this problem? It should target `master` branch, of course. Thanks! 👍

cllns:
@wjdhamilton Want to take this one? If not, I can do it. Feel free to ping me in the [Hanami Gitter chat](https://gitter.im/hanami/chat) if you need any help :)

wjdhamilton:
Yes, I’ll take this one. I will probably have to ask you a few questions first but I’ll get on it later this week.

tejasbubane:
I guess this can be closed.

jodosha:
Fixed by #325

Issue #250, opened 2015-10-14
Improve error message for `lotus db drop`


jodosha:
Steps to reproduce.

```shell
% lotus new bookshelf --database=postgres && cd bookshelf && bundle
% bundle exec lotus db prepare
```

Open another shell and left open the PG client

```shell
% psql bookshelf_development
```

Get back to the initial shell:

```shell
% bundle exec lotus db drop
/Users/luca/.gem/ruby/2.2.3/gems/lotus-model-0.5.0/lib/lotus/model/migrator/postgres_adapter.rb:51:in `block in drop': Lotus::Model::MigrationError (Lotus::Model::MigrationError)
        from /Users/luca/.rubies/ruby-2.2.3/lib/ruby/2.2.0/open3.rb:199:in `popen_run'
        from /Users/luca/.rubies/ruby-2.2.3/lib/ruby/2.2.0/open3.rb:93:in `popen3'
        from /Users/luca/.gem/ruby/2.2.3/gems/lotus-model-0.5.0/lib/lotus/model/migrator/postgres_adapter.rb:39:in `drop'
        from /Users/luca/.gem/ruby/2.2.3/gems/lotus-model-0.5.0/lib/lotus/model/migrator.rb:112:in `drop'
        from /Users/luca/.gem/ruby/2.2.3/gems/lotusrb-0.5.0/lib/lotus/commands/db/drop.rb:9:in `start'
        from /Users/luca/.gem/ruby/2.2.3/gems/lotusrb-0.5.0/lib/lotus/commands/db.rb:46:in `drop'
        from /Users/luca/.gem/ruby/2.2.3/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from /Users/luca/.gem/ruby/2.2.3/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from /Users/luca/.gem/ruby/2.2.3/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from /Users/luca/.gem/ruby/2.2.3/gems/thor-0.19.1/lib/thor/invocation.rb:115:in `invoke'
        from /Users/luca/.gem/ruby/2.2.3/gems/thor-0.19.1/lib/thor.rb:235:in `block in subcommand'
        from /Users/luca/.gem/ruby/2.2.3/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
[...]

PR #325, opened 2016-06-28
Updated error message as per #322

wjdhamilton:
It would probably be better in the long run if a gist was created in the hanami repo or a blog post in hanamirb to properly explain the console issue, but I think this will suffice for now given that the ROM upgrade is on its way.

### DIFF
--- a/lib/hanami/model/migrator/postgres_adapter.rb
+++ b/lib/hanami/model/migrator/postgres_adapter.rb
@@ -29,7 +29,10 @@ module Hanami
 
           call_db_command('createdb') do |error_message|
             message = if error_message.match(/already exists/)
-              "createdb: database creation failed. There is 1 other session using the database."
+              "createdb: database creation failed.If the database exists, \ 
+               then its console may be open. See this issue for more details:\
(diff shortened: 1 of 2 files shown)


cllns:
This addresses #322 (in title here but putting in this comment so it's referenced in that issue)

Sorry for the delay here. I don't think we should link to #250 in the error message since it's only tangentially related (it's for `dropdb` not `createdb`). Seems potentially more confusing than helpful.

Can we change the error message to just `createdb: database creation failed. If the database exists, then its console may be open.` ? I think that's clear enough. What do you think, @wjdhamilton?

wjdhamilton:
Hi Sean, yes. Um...only one small problem which is that I mistakenly deleted my branch in the intervening period. One of those times I really should have read the warning message. Anyway, I will effect those changes in a new repo and get the pull request in.

cllns:
No problem! It's a small enough fix and GitHub cached the diff (on the Changes page) so you can re-do it from that.
Thanks!

jodosha:
@wjdhamilton I merged it, thanks!

PR #329, opened 2016-07-11
adapter configuration for 0.6.x 

kucaahbe:
Added possibility to pass any additional adapter configuration for 0.6.x version. Please, let me know if any other things I need to do in order to this PR could be submitted.

### DIFF
--- a/lib/hanami/model/migrator.rb
+++ b/lib/hanami/model/migrator.rb
@@ -290,7 +290,8 @@ module Hanami
       # @api private
       def self.connection
         Sequel.connect(
-          configuration.adapter.uri
+          configuration.adapter.uri,
+          configuration.adapter.options
         )
       rescue Sequel::AdapterNotFound
         raise MigrationError.new("Current adapter (#{ configuration.adapter.type }) doesn't support SQL database operations.")


joneslee85:
@kucaahbe it'd be nice if you could provide usecases for this change. Besides, this version is too old.

kucaahbe:
@joneslee85 for instance to pass any option in order to configure Sequel, e.g. `sslmode`

AlfonsoUceda:
@kucaahbe why modify this version, can't you update to the last one? maybe the new model version can use that, @jodosha can you confirm us this?

kucaahbe:
@AlfonsoUceda for the moment when PR was created there wasn't later version (as far as I remember), only master branch. So, assuming this change could be useful, don't break anything and could be merged to future versions I've created PR for 0.6.x. Let me know how can I improve PR and against which branch it should be submitted, thanks!

jodosha:
@kucaahbe Hi, what's your use case? IIRC you can express extra options for Sequel via query string: eg. `postgres://localhost/bookshelf_development?extra_option=here`. Can you please try if this works for you? Thanks.

AlfonsoUceda:
Sorry, but I don't agree create new tags for an old version when we are under version 1.0, I think the only purpose for creating a new tag for an old version is for security reasons.

@kucaahbe could you check what @jodosha said?

kucaahbe:
@jodosha @AlfonsoUceda proposed solution tend to work, but worth to mention that it probably would be nice to have database configuration separation, where some part of configuration is local and some is global, (e.g. encoding option). However if it's too hard (so much use-cases to learn and tags to create, just kidding) ), than don't see a reason to keep this pr open, thanks!

AlfonsoUceda:
@kucaahbe thanks, but this can be proposed in master instead this one, we can discuss about that ;)

PR #334, opened 2016-09-19
New Engine (ROM)

### LINKED ISSUE #217: Error on entity dirty tracking
I got an unexpected "error" in my entity using dirty tracking. If I change the includes order in my entity the `changed?` and `changed_attributes` does not work perfectly.
# For this includes order

``` ruby
Lotus::Model::VERSION # => "0.4.1"

class Pet
  include Lotus::Entity
  include Lotus::Entity::DirtyTracking
  include Lotus::Validations

  attribute :name, type: String
end

pet = PetRepository.first
pet.changed?           # => false
pet.name = 'name'      # => "name"
pet.changed?           # => false
pet.changed_attributes # => {}
```
# Inverting Validations and DirtyTracking

``` ruby
class Pet
  include Lotus::Entity
  include Lotus::Validations
  include Lotus::Entity::DirtyTracking

  attribute :name, type: String
end

pet = PetRepository.first
pet.changed?           # => false
pet.name = 'name'      # => "name"
pet.changed?           # => true
pet.changed_attributes # => {:name=>"Bacon"}
```

### LINKED ISSUE #237: Repository does not respect default values on persisted entities

### LINKED ISSUE #245: Path glob to mapper files
Hi!

Organize my mapping files I separate mapping in many files that makes sense for me. So I created a main mapping file that it got all path, as following:

``` ruby
﻿Lotus::Model.configure do
  adapter type: :sql, ...

  mapping 'mapping/users'
  mapping 'mapping/tasks'
  mapping 'mapping/scheduled'
end
```

It works very well, but it more redundancy for each mapping I will need putting these files in main mapping file.

My suggestion if will be possible use a glob path in main file. For instance, my file could be more elegant as following:

``` ruby
﻿Lotus::Model.configure do
  adapter type: :sql, ...

  mapping 'mapping/*'
end
```

I saw that it is not possible at the moment, because the `Config::Model`(https://github.com/lotus/model/blob/master/lib/lotus/model/config/mapper.rb#L34) try get a realpath.

If guys agree with me, I will implement it refactoring the code.

### LINKED ISSUE #272: Booleans can't be dumped in repository
```
undefined method `dump' for TrueClass:Class
```

Given the mapping

``` Ruby
collection :foos do
  entity     Foo
  repository FooRepository
  attribute :bar, TrueClass
end
```

And the creation

``` Ruby
@foo = Foo.new( params[ :foo ] ) #form_for :foo
FooRepository.create( @foo )
```

### LINKED ISSUE #273: Empty values can't be coerced
When a form is submitted and the value is empty (but present as key because it was in the form).

Coercing

```
'' Can't be coerced into Date
```

### LINKED ISSUE #278: Raise error when :id isn't defined in a mapping since it's required by the repositories
If you set a custom field as primary key using the identity method in a collection mapping, the update action on the repository doesn't work because it checks explicitly for the id of the entity. It's defined here:

https://github.com/lotus/model/blob/9801dd781dc107aa0888430b220c0293d3eff258/lib/lotus/repository.rb#L830-L837

### LINKED ISSUE #279: An update action should not include the ID
If you define an alternate primary key in a mapping with 'identity' this is included in the fields to be updated. Consider this query arriving in an SQL server, it was generated with Lotus model of course:

```
UPDATE [CLIENTS] SET [CODE] = N'0001      ', [DIREC1] = N' Le adress', [DIREC2] = N'ALOHA' WHERE ([CODE] = N'0001      ')
```

A validation in the DB complain that I cannot change the primary key. I would assume it makes sense not to send the primary key to be modified or only send the keys that have changed, e.g:

```
UPDATE [CLIENTS] SET [DIREC1] = N' Le adress', [DIREC2] = N'ALOHA' WHERE ([CODE] = N'0001      ')
```

### LINKED ISSUE #285: Better error for 'deadlock; recursive locking'  - Memory Adapter
A week ago or so I came across a problem where, inside a `query` block, I called another method which called `query`, using the `memory_adapter`. I got this error:

```
     ThreadError:
       deadlock; recursive locking
```

This is because [`query`](https://github.com/lotus/model/blob/master/lib/lotus/model/adapters/memory_adapter.rb#L125) uses a mutex to `synchronize` the query, so it can't recursively call another `query`.

I fixed it by storing the results of the first method that called `query` into a variable, then using that variable inside the `query`.

I think this only affects `memory_adapter`.

Should we catch `ThreadError` and throw a new `Lotus::Error`, with a more helpful error message? Basically just tell the user what happened, and how they can avoid it.

At the very least, if we don't want that, we can close this Issue and it'll show up in search results when people have this problem :)

(Thanks to @angelbotto for reminding me of this issue, in the chat)

### LINKED ISSUE #291: Make repositories to work with data

### LINKED ISSUE #294: Make entities immutable
**Please note that this is a proposal. We want to discuss first and evaluate if this will be implemented and included in the next release.**

---

Mutable state is known to be difficult to handle and dangerous for concurrency.

This is a proposal to remove the following features from `Hanami::Entity`:
- Setters
- `#update`
- `Hanami::Entity::DirtyTracking`

Besides to these changes, we should probably amend `Repository#update` signature by accepting the id and the data.

If we follow the principle _"data in, entities out"_ (described by https://github.com/hanami/model/issues/291), entities are only the output of database operations.

Please check one by one the following CRUD examples and spot the points where we need to mutate an entity.
### Create

``` ruby
def call(params)
  book = BookRepository.new.create(params[:book])
  redirect_to routes.books_url(id: book.id)
end
```
### Read

``` ruby
expose :book

def call(params)
  @book = BookRepository.new.find(params[:id])
end
```

The associated view/template only needs getters to present data.
### Update

``` ruby
def call(params)
  book = BookRepository.new.update(params[:id], params[:book])
  redirect_to routes.books_url(id: book.id)
end
```
### Delete

``` ruby
def call(params)
  book = BookRepository.new.destroy(params[:id])
  redirect_to routes.books_url
end
```

For default CRUD operations we don't need to mutate the state of an entity. I'm looking for feedback to understand where mutation (and dirty tracking) can be still useful.

---

/cc @hanami/core-team @hanami/contributors

jodosha:
## New Engine

Use `rom` as low-level engine for `hanami-model`. Introduce new features (automapping for SQL databases, experimental associations), remove old (dirty tracking). Fix old bugs.
### Example

Before to dive into the details here's how it looks the basic example:

``` ruby
require 'hanami/model'

class User
  include Hanami::Entity
  # no more attributes here
end

class UserRepository < Hanami::Repository # inherit instead of include
end

Hanami::Model.configure do
  adapter :sql, ENV['DATABASE_URL']
  # No more mapping
end.load!

user       = repository.create(name: 'Luca') # It works with data as primary input type
user.id # => 1

user       = repository.update(user.id, age: 34) # It accepts the primary key and the data
user.age # => 34

repository.delete(user.id) # It accepts the primary key
```

This **reduces drastically the boilerplate** to setup an entity and a repository pair.
### Repositories

Keeping the good work from #299, repositories now expose their CRUD operations at the instance level, not at the class level.

Data (hashes) is the preferred input for CRUD operations. In past we used to pass entities, while this is still possible, it’s suggested to use just data.

Repositories can access all the relations with instance methods. For instance, from the `BookRepository` we can access `users` relation (database table) with `#users`.

Repositories are able to _auto infer_ the database schema and use the correct Ruby types. Advanced PostgreSQL types like UUID, Array, JSON(B) and Money are handled natively.

Repositories can also _auto map_ database columns with entities attributes. In case there is a mismatch between the names, we can define a custom mapping. **NOTE:** manual mapping is optional and should be used only with legacy databases.

``` ruby
# Basic case
end

# Advanced case for legacy databases
class OperatorRepository < Hanami::Repository
  # By convention we expect it to work with :operators,
  # but instead we manually define the name
  self.relation = :t_operator

  # Define a mapping between each entity attribute to the corresponding database column
  mapping do
    attribute :id,   from: :i_operator_id
    attribute :name, from: :s_name
    # …
  end
end
```

Custom queries can be defined with an intuitive Ruby syntax:

``` ruby
  def by_name(name)
    users.where(name: name).as(:entity)
  end
end
```

### Entities

Entities are now **immutable**.

#### Automatic Schema

With SQL databases, entities get attributes from table schema.

As example, image to have a `users` table with `id` (int), `name` (string), `email` (string), `age` (int), `created_at` (time), and `updated_at` (time).

```ruby
class User < Hanami::Entity
end

end

user = User.new(name: "Luca", foo: "bar")
user.name # => "Luca"
user.age # => nil
user.foo # => NoMethodError
```

It ignores `foo` because it isn't part of the schema. `age` instead is `nil` because we haven't initialized it.

There is also a **new feature** for entities which **enforces schema types**. When an entity is initialized with one or more wrong attributes it raises an error.

```ruby
class User < Hanami::Entity
end

end

User.new(age: "foo") # => ArgumentError
```

#### Custom Schema

If developers want to customize the set of attributes of an entity or specify the level of strictness of the type coercion, they can build a **custom schema**.

```ruby
class User < Hanami::Entity
  attributes do
    attribute :id, Types::Int
    # ...
  end
end

end
```

For a complete list of the options, please check: http://dry-rb.org/gems/dry-types/built-in-types/

Please note that `.attributes` changed its public API.

### Associations (experimental)

We introduced experimental support for associations (only `has_many` at this time).

When we define an association, it does NOT add any method to the public interface. We can access the association internally via `#assoc` and define only the explicit operations that we need for our model domain.

``` ruby
class AuthorRepository < Hanami::Repository
  associations do
    has_many :books
  end

  def find_with_books(id)
    aggregate(:books).where(authors__id: id).as(Author).one
  end

  def books_for(author)
    assoc(:books, author)
  end

  def add_book(author, data)
    assoc(:books, author).add(data)
  end

  def remove_book(author, id)
    assoc(:books, author).remove(id)
  end

  def delete_books(author)
    assoc(:books, author).delete
  end

  def delete_on_sales_books(author)
    assoc(:books, author).where(on_sale: true).delete
  end

  def books_count(author)
    assoc(:books, author).count
  end

  def on_sales_books_count(author)
    assoc(:books, author).where(on_sale: true).count
  end

  def find_book(author, id)
    book_for(author, id).one
  end

  def book_exists?(author, id)
    book_for(author, id).exists?
  end

  private

  def book_for(author, id)
    assoc(:books, author).where(id: id)
  end
end
```

---

A special thanks goes to @solnic, @flash-gordon, @AMHOL and @timriley for ROM and their tireless support provided during these months of development. You rock! 💚 

---

Closes #217 
Closes #237 
Closes #245 
Closes #272 
Closes #273 
Closes #278 
Closes #279 
Closes #285 
Closes #291 
Closes #294 

---

/cc @hanami/core-team @hanami/contributors


lib/hanami/entity.rb
```diff
@@ -61,238 +51,27 @@ module Hanami
...
-    # @since 0.1.0
-    #
-    # @see .attributes
-    def initialize(attributes = {})
-      attributes.each do |k, v|
-        setter = "#{ k }="
-        public_send(setter, v) if respond_to?(setter)
-      end
+    def method_missing(m)
+      attributes.fetch(m, nil)
```
AMHOL: Not sure about responding to everything here, would be better to check for the key in attributes or read from the relation schema to define the interface, also need to define `#respond_to_missing?(method, _include_private = false)` I think

lib/hanami/entity.rb
```diff
@@ -61,238 +51,27 @@ module Hanami
...
-    # Overrides the equality Ruby operator
-    #
-    # Two entities are considered equal if they are instances of the same class
-    # and if they have the same #id.
-    #
-    # @since 0.1.0
     def ==(other)
       self.class == other.class &&
-         self.id == other.id
+        id == other.id
```
AMHOL: Should equality check just be by ID? Could read from the relation schema for the attribute definitions and use [dry-equalizer](http://github.com/dry-rb/dry-equalizer) for this

lib/hanami/model/association.rb
```diff
@@ -0,0 +1,23 @@
...
+      # Instantiate an association
+      #
+      # @since x.x.x
+      # @api private
+      def self.new(repository, target, subject)
+        case repository.root.associations[target]
+        when ROM::SQL::Association::OneToMany then Associations::HasMany
+        else
+          raise 'unsupported association'
+        end.new(repository, repository.root.name.to_sym, target, subject)
```
AMHOL: I'd go with something like this here rather than overriding new and returning an instance of a different class (could catch somebody out)

```ruby
module Hanami
  module Model
    class Error < StandardError
      UnsupportedAssociation = Class.new(Error)
    end

    class Association
      MAPPING = {
        ROM::SQL::Association::OneToMany => Associations::HasMany
      }

      attr_accessor :repository, :target, :subject
      private :repository, :target, :subject

      def initialize(repository, target, subject)
        @repository, @target, @subject = repository, target, subject
      end

      def association
        association_class.new(repository, repository.root.name.to_sym, target, subject)
      end

      private

      def association_class
        MAPPING.fetch(association_adapter_class) do
          raise(
            UnsupportedAssociation,
            "unsupported association #{association_adapter_class}"
          )
        end
      end

      def association_adapter_class
        repository.root.associations[target]
      end
    end
  end
end
```

lib/hanami/model/configuration.rb
```diff
@@ -225,50 +43,24 @@ def root
         Hanami.respond_to?(:root) ? Hanami.root : Pathname.pwd
```
AMHOL: Defaulting to `pwd` here seems strange, wouldn't it be better to just raise?

lib/hanami/model/migrator/adapter.rb
```diff
@@ -38,15 +40,30 @@ def self.for(connection)
...
+          end.new(connection, configuration)
+        end
+
+        class << self
+          private
+
+          # @since x.x.x
+          # @api private
+          def connection_for(configuration)
+            Sequel.connect(
```
AMHOL: You can do this in rom by creating a configuration without creating a container from it

lib/hanami/model/migrator/connection.rb
```diff
@@ -81,31 +85,39 @@ def password
...
-          @uri ||= URI.parse(adapter_connection.uri.sub('jdbc:', ''))
+          @uri ||= URI.parse(raw.uri.sub('jdbc:', ''))
+        end
+
+        # Return the database table for the given name
+        #
+        # @since x.x.x
+        # @api private
+        def table(name)
+          raw[name] if raw.tables.include?(name)
```
AMHOL: Is this just a hash? Will do the same without the if statement if it is

lib/hanami/model/sql.rb
```diff
@@ -0,0 +1,158 @@
...
+      #         # ...
+      #       end
+      #     end
+      #
+      #     down do
+      #       drop_table :itmes
+      #       execute 'DROP TYPE inventory_item'
+      #     end
+      #   end
+      def self.literal(string)
```
AMHOL: This could be confused with `Sequel::Database#literal`

lib/hanami/entity.rb
jodosha: @AMHOL Yeah, absolutely. Entity is reduced to dumb interface for now.

lib/hanami/entity.rb
jodosha: Nope, we should check class too. Otherwise `Book.new(id: 1) == User.new(id: 1)` would return `true`.

lib/hanami/model/association.rb
jodosha: @AMHOL This looks better, but I would prefer a _fail fast_ strategy. I'd move `association_class.new` into `#initialize`, so it would blow instantly when try to instantiate an unknown type of association. 😉

lib/hanami/model/migrator/connection.rb
jodosha: @AMHOL Nope, this is a Sequel object which would raise an exception for unknown tables. 

We can't let this to fail because the caller is usually an object that backs CLI commands. For instance `hanami db version`. If there isn't migrations table we won't raise an exception, but just tell the user that we're still at version 0.

lib/hanami/entity.rb
pascalbetz: I don't like the respond to everything aproach either. I prefer fail-fast.

lib/hanami/model/sql.rb
jodosha: @AMHOL IMO, naming it `.lit` isn't communicative enough like `.literal`.

lib/hanami/entity.rb
pascalbetz: Should `hash` be implemented as well?

lib/hanami/model/adapters/sql_adapter.rb
```diff
@@ -1,296 +0,0 @@
...
-        #
-        # @api private
-        # @since 0.1.0
-        def _collection(name)
-          Sql::Collection.new(@connection[name], _mapped_collection(name))
-        end
-      end
-    end
-  end
-end
```
pascalbetz: Amazing how much code was deleted!

lib/hanami/model/association.rb
pascalbetz: Agree with @AMHOL here. Returning a different class is dangerous.

lib/hanami/model/entity_name.rb
```diff
@@ -0,0 +1,25 @@
...
+      # @return [String] the entity name
+      #
+      # @since x.x.x
+      # @api private
+      def self.new(name)
+        name.sub(SUFFIX, '')
+      end
+    end
+  end
+end
```
pascalbetz: I like the encapsulation.

lib/hanami/entity.rb
```diff
@@ -56,8 +56,12 @@ def initialize(attributes = nil)
       freeze
     end
 
+    def id
+      attributes.fetch(:id, nil)
+    end
+
     def method_missing(m)
-      attributes.fetch(m, nil)
+      attributes.fetch(m) { super }
```
pascalbetz: I like this better! 👍

lib/hanami/model/configuration.rb
mereghost: For me it seems sensible if hanami-model isn't been used on a Hanami project. Raising an exception in this case would be a bit baffling, IMO.

README.md
```diff
@@ -232,14 +199,11 @@ There is a better way:
...
 
-class ArticleRepository
-  include Hanami::Repository
-
-  def self.most_recent_by_author(author, limit = 8)
-    query do
-      where(author_id: author.id).
-        order(:published_at)
-    end.limit(limit)
+class ArticleRepository < Hanami::Repository
```
jodosha: TBH, because of ROM. It's unconvenient to make `ROM::Repository::Root` work without inheritance.

lib/hanami/model/configuration.rb
jodosha: This should stay `pwd` because in configuration we use to do:

```ruby
Hanami::Model.configure do
  # ...
  migrations 'db/migrations'
  schema     'db/schema.sql'
end
```

As you can see, that are relative paths, that joined with `pwd` return the correct directory/file.

lib/hanami/model/migrator/adapter.rb
jodosha: @AMHOL Internally we're still using Sequel directly for migrations.

lib/hanami/model/association.rb
jodosha: For now I just changed it to `Association.build` (see https://github.com/hanami/model/pull/334/commits/480d1f48a2e1d9297ae5cc514c6cca3b93eb263b). This is still experimental code that needs to be revised later. It will get a better shape later 😉

lib/hanami/entity.rb
AMHOL: @jodosha sorry I meant should it act like a value object and check equality using all attributes

lib/hanami/model/association.rb
AMHOL: Cool :)

lib/hanami/model/configuration.rb
AMHOL: Ahh OK :+1:

lib/hanami/model/migrator/connection.rb
AMHOL: OK, thanks for the explanation :smile:

lib/hanami/model/sql.rb
AMHOL: Fair enough, I guess my comment is only true of people experienced with Sequel anyway

README.md
AMHOL: The idea in ROM is that using inheritance indicates that the class should have one responsibility, i.e. we're inheriting repository, so this class should ONLY act as a repository, discouraging people from expanding the interface with mixins

lib/hanami/model/sql.rb
jodosha: Probably yes 😉

lib/hanami/model/association.rb
pascalbetz: 👍

lib/hanami/entity.rb
```diff
@@ -51,33 +51,81 @@ module Hanami
...
     private
 
+    # @since 0.1.0
+    # @api private
     attr_reader :attributes
 
+    # @since x.x.x
+    # @api private
     def respond_to_missing?(name, _include_all)
       attributes.key?(name)
```
jodosha: Why delegate to `super`? The only source of dynamic method dispatch is `attributes`.

According to [Ruby code docs](https://ruby-doc.org/core/Object.html#method-i-respond_to_missing-3F), `include_all` doesn't have a default value.

lib/hanami/entity.rb
```diff
@@ -61,238 +51,35 @@ module Hanami
...
-    #   User = Struct.new(:id, :name) do
-    #     include Hanami::Entity
-    #   end
-    def self.included(base)
-      base.class_eval do
-        extend ClassMethods
-        attributes :id
-      end
+    def initialize(attributes = nil)
+      @attributes = Utils::Hash.new((attributes || {}).dup).symbolize!
```
pascalbetz: the current code covers 

`Something.new(nil)` 

as well.

lib/hanami/entity.rb
AMHOL: I just do it as a general rule so I don't introduce bugs when something is a subclass or could be used as a mixin. Must have confused `include_all` with the `respond_to?` method argument lol

lib/hanami/entity.rb
jodosha: Yes, that ensures that an entity instance will always have an encapsulated hash.

.travis.yml
```diff
@@ -1,18 +1,23 @@
 language: ruby
 sudo: false
 cache: bundler
-script: 'bundle exec rake test:coverage --trace'
+script: 'bundle exec rubocop && bundle exec rake test:coverage --trace'
```
cllns: Glad to see we're using rubocop btw :)

lib/hanami/model/associations/dsl.rb
```diff
@@ -0,0 +1,29 @@
+module Hanami
+  module Model
+    module Associations
+      # Auto-infer relations linked to repository's associations
+      #
+      # @since x.x.x
+      # @api private
+      class Dsl
```
cllns: nitpicky but can this be `DSL`? It's an acronym so we should capitalize if we can

lib/hanami/model/migrator/mysql_adapter.rb
```diff
@@ -1,70 +1,88 @@
...
-          new_connection(global: true).run %(CREATE DATABASE #{ database };)
+          new_connection(global: true).run %(CREATE DATABASE #{database};)
         rescue Sequel::DatabaseError => e
-          message = if e.message.match(/database exists/)
-            "Database creation failed. There is 1 other session using the database"
-          else
-            e.message
-          end
+          message = if e.message.match(/database exists/) # rubocop:disable Performance/RedundantMatch
+                      'Database creation failed. There is 1 other session using the database'
```
cllns: Can we update this message? I don't think it's necessarily accurate that `There is 1 other session using the database`. https://github.com/hanami/model/pull/325/files

lib/hanami/model/sql/consoles/mysql.rb
```diff
@@ -0,0 +1,65 @@
...
+          # @since x.x.x
+          # @api private
+          def connection_string
+            str = 'mysql'
+            str << host
+            str << database
+            str << port if port
+            str << username if username
+            str << password if password
+            str
```
cllns: string concatenation feels kind of strange to me, what about:
```
[
  'mysql',
  host,
  database,
  port,
  username,
  password
].join
```
or on a single line even

.travis.yml
```diff
@@ -1,18 +1,23 @@
...
   - jruby-9.0.5.0
   - ruby-head
   - jruby-head
+env:
+  - DB=sqlite
+  - DB=postgresql
+  - DB=mysql
 
 addons:
   postgresql: '9.4'
```
cllns: FYI, we could add 9.5, if we want to use Travis's `trusty` build (which is is beta). `precise` is the default and doesn't support 9.5 yet though
https://docs.travis-ci.com/user/database-setup/#PostgreSQL

lib/hanami/model/sql/consoles/mysql.rb
pascalbetz: @cllns it's not the same (note the statement modifiers/conditionals).

But I still agree with your solution:-)

`String#<<` will not work with `frozen_string_literal`

So it should be

```ruby
parts = []
parts << "something"
parts << "else" if condition
parts.join # is the connection string concatenated with empty string?
```

lib/hanami/entity.rb
jodosha: @AMHOL I'm not sure about this. For value objects the "identity" is made by its values, whereas for entities the identity is an unique identifier, in this case the primary key from the database.

README.md
jodosha: @AMHOL Thanks for the explanation. But this would be an exception in the entire framework where we always have mixins to include.

lib/hanami/entity.rb
pascalbetz: @jodosha so following should be `true`?
```ruby
Thing.new(id: 12, name: 'Another thing') == Thing.new(id:12, name: 'A thing')
```

I think the problem lies within the multiple meanings of `equal`

* Same state
* Same entity
* And the combination

From a user point of view i'd expect the above code sample toe be `false`

.travis.yml
jodosha: @cllns For now I'd keep things simple and avoid yet another dep. Things may change after 1.0. 😉

lib/hanami/model/migrator/mysql_adapter.rb
jodosha: Fixed with ef50623 ;)

lib/hanami/model/sql/consoles/mysql.rb
jodosha: See cc3067b 😉

.travis.yml
```diff
@@ -1,18 +1,23 @@
...
 
 matrix:
+  fast_finish: true
```
jodosha: Removed by db8355f

lib/hanami/repository.rb
```diff
@@ -118,769 +106,322 @@ module Hanami
...
+    # Return all the records for the relation
+    #
+    # @return [Array<Hanami::Entity>] all the entities
+    #
+    # @since x.x.x
+    #
+    # @example
+    #   UserRepository.new.all
+    def all
+      root.as(:entity)
```
AMHOL: Should `#to_a` be called here to materialize the results and prevent chaining?

lib/hanami/repository.rb
```diff
@@ -118,769 +106,322 @@ module Hanami
...
-          entity.respond_to?("#{ timestamp }=")
-      end
+    # Deletes all the records from the relation
+    #
+    # @since x.x.x
+    #
+    # @example
+    #   UserRepository.new.clear
+    def clear
+      root.delete
```
AMHOL: Should `#as` and `#to_a` be called here to materialize the results and prevent chaining?

lib/hanami/repository.rb
cllns: I think I'd prefer to keep it chainable, in terms of being friendly to users. Don't feel super strongly though.

I could see someone having `UserRepository.new.all` then adding a where, like:`UserRepository.new.all.where(first_name: "Andy")`. The `all` is unnecessary, but it also doesn't really hurt anything either. ¯\\\_(ツ)\_/¯ 
Otherwise they'll get `NoMethodError: undefined method 'where' for []:Array`. If users want to materialize the result, they can call `to_a` themselves. It'd be surprising to me if `all` did that.

AMHOL: @cllns I mean to be in keeping with https://github.com/hanami/model/blob/af211f3259848e31e9dd64a33a7f70e37f1da9ba/lib/hanami/repository.rb#L54-L108

cllns:
If anyone wants to try this out in a project, here's what I added to my `Gemfile`, after `gem 'hanami', '~> 0.8'`:

```
gem 'hanami-model', github: 'hanami/model', branch: 'new-engine'
gem 'hanami-validations', github: 'hanami/validations'
gem 'dry-validation', github: 'dry-rb/dry-validation', branch: 'master'
gem 'dry-types', require: false, github: 'dry-rb/dry-types' # FIXME: this is needed until they release 0.9
gem 'dry-struct', github: 'dry-rb/dry-struct', branch: 'master'

gem 'hanami-utils', '~> 0.8', require: false, github: 'hanami/utils',    branch: 'master'
gem 'rom-sql', '~> 0.8', require: false, github: 'jodosha/rom-sql', branch: 'hanami-model-integration'
```

Some other settings need to be changed too. Will update this comment with more info once I figure it out :)

README.md
```diff
@@ -256,253 +220,32 @@ This is a **huge improvement**, because:
...
-    attribute :id,    Integer
-    attribute :name,  String
-    attribute :price, Integer
+  mapping do
+    attribute :id,   from: :i_user_id
+    attribute :name, from: :s_name
+    attribute :age,  from: :i_age
   end
 end
-```
```
cllns: Need these ``` to close the codeblock (see rendering here https://github.com/hanami/model/blob/af211f3259848e31e9dd64a33a7f70e37f1da9ba/README.md)

README.md
```diff
@@ -192,27 +161,25 @@ This architecture has several advantages:
...
+  * `#all`              - Fetch all the entities from the relation
+  * `#find`             - Fetch an entity from the relation by primary key
+  * `#first`            - Fetch the first entity from the relation
+  * `#last`             - Fetch the last entity from the relation
+  * `#clear`            - Delete all the records from the relation
 
-**A collection is a homogenous set of records.**
-It corresponds to a table for a SQL database or to a MongoDB collection.
+**A relation is a homogenous set of records.**
+It corresponds to a table for a SQL database or to a MongoDB relation.
```
cllns: It looks like MongoDB refers to them as collections, so this line shouldn't change.

hanami-model.gemspec
```diff
@@ -8,8 +8,8 @@ Gem::Specification.new do |spec|
   spec.version       = Hanami::Model::VERSION
   spec.authors       = ['Luca Guidi', 'Trung Lê', 'Alfonso Uceda']
   spec.email         = ['me@lucaguidi.com', 'trung.le@ruby-journal.com', 'uceda73@gmail.com']
-  spec.summary       = %q{A persistence layer for Hanami}
-  spec.description   = %q{A persistence framework with entities, repositories, data mapper and query objects}
+  spec.summary       = 'A persistence layer for Hanami'
+  spec.description   = 'A persistence framework with entities, repositories, data mapper and query objects'
```
cllns: Do we want to keep 'data mapper here'? It's only needed for legacy cases

README.md
```diff
@@ -117,57 +105,38 @@ require 'hanami/model'
...
+    @attributes
   end
+  alias to_hash to_h
 end
 ```
 
 **Hanami::Model** ships `Hanami::Entity` for developers's convenience.
 
-**Hanami::Model** depends on a narrow and well-defined interface for an Entity - `#id`, `#id=`, `#initialize(attributes={})`.
+**Hanami::Model** depends on a narrow and well-defined interface for an Entity - `#initialize(attributes={})`, `#to_h`.
```
cllns: Is the resulting hash expected to have a key called `id`? If so, we might want to mention it here.

lib/hanami/model/mapping.rb
```diff
@@ -1,47 +1,29 @@
...
-      # @since 0.1.0
-      class UnmappedCollectionError < Hanami::Model::Error
-        def initialize(name)
-          super("Cannot find collection: #{ name }")
-        end
+    class Mapping
+      def initialize(&blk)
+        @attributes = {}
+        instance_eval(&blk)
+        @processor = @attributes.empty? ? ::Hash : Transproc(:rename_keys, @attributes)
```
cllns: I get this deprecation notice:
```
Transproc() is deprecated and will be removed in 1.0.0.
Define your own function registry using Transproc::Registry extension
```

Here's the diff to fix it 
```diff
     class Mapping
+      extend Transproc::Registry
+
       def initialize(&blk)
         @attributes = {}
         instance_eval(&blk)
-        @processor = @attributes.empty? ? ::Hash : Transproc(:rename_keys, @attributes)
+        @processor = @attributes.empty? ? ::Hash : Transproc[:rename_keys, @attributes]
       end
```

AMHOL: I'd go with something more like:

```ruby
class Mapping
  Transformer = Module.new do
    extend Transproc::Registry
    import :rename_keys, from: Transproc::HashTransformations
  end

  def initialize(&blk)
    @attributes = {}
    instance_eval(&blk)
    @processor = @attributes.empty? ? ::Hash : Transformer[
      :rename_keys,
      @attributes
    ]
  end
  # ...
```

cllns: I'm against multi-line ternaries but other than that, I trust you :)

lib/hanami/repository.rb
cllns: ah, I see!

solnic: Once folks can do `repo.all.where(...).limit(...).joins(...)` in their controllers, we're back on Rails. Just my 2 cents ;)

lib/hanami/model/mapping.rb
```diff
@@ -1,47 +1,29 @@
...
-      # It gets raised when the application tries to access to a existing
-      # repository.
-      #
-      # @since 0.2.0
-      class RepositoryNotFound < Hanami::Model::Error
-        def initialize(name)
-          super("Cannot find class for repository: #{ name }")
-        end
+      def register_as(name)
+      end
```
solnic: What's the purpose of this no-op method?

lib/hanami/entity.rb
```diff
@@ -61,238 +51,83 @@ module Hanami
...
+    # @return [FalseClass,TrueClass] the result of the check
     #
-    #   user = User.new(id: 23, name: 'Luca')
-    #   user.attribute_names # #<Set: {:id, :name}>
-    def attribute_names
-      self.class.attributes
+    # @since 0.1.0
+    def ==(other)
+      self.class == other.class &&
+        id == other.id
```
beauby: In what context is this equality expected to be used?

lib/hanami/entity.rb
```diff
@@ -61,238 +51,83 @@ module Hanami
...
-    #   user = User.new(name: 'Lucca')
-    #   user.update(name: 'Luca')
-    #   user.name # => 'Luca'
-    def update(attributes={})
-      attributes.each do |attribute, value|
-        public_send("#{attribute}=", value)
-      end
+    # @since 0.1.0
+    def to_h
+      attributes.deep_dup.to_h
```
beauby: Are attributes containing associations here? If so this would blow up in case of a `User` with a symmetrical friendship association.

lib/hanami/entity.rb
mereghost: From what I get, this is Entity equality as defined by DDD, so it should be used whenever you need to check if a pair of entities are the same.

beauby: I'm not really familiar with DDD (any interesting read?), but what's the value in having a stale entity A be equal to itself after modification? (`A = UserRepository.create(...); B = UserRepository.modifyEmail(A.id, ...); A == B # => true`)

beauby:
I think most of the heavy lifting has been done here, the last (and one of the most important) point is to choose the right API. Most of it should be set by ROM, which is currently the case, except for the associations (whose current API is "access associations from the repository through the `assoc` method (one parameter of which currently seems kinda redundant), and then use the methods defined on the association itself (currently only `has_many`)). Is there a place to discuss those?

lib/hanami/entity.rb
solnic: @beauby there are no circular associations in rom (thus hanami-model). we're dealing with aggregates, so only root objects know about their children, and never the other way around.

lib/hanami/model.rb
```diff
@@ -1,172 +1,84 @@
...
-    #   end
-    #
-    #   Hanami::Model.configuration.adapter_config # => nil
-    #   MyApp::Model.configuration.adapter_config # => #<Hanami::Model::Config::Adapter:0x007ff0ff0244f8 @type=:memory, @uri="memory://localhost", @class_name="MemoryAdapter">
-    def self.duplicate(mod, &blk)
-      dupe.tap do |duplicated|
-        mod.module_eval %{
-          Entity = Hanami::Entity.dup
-          Repository = Hanami::Repository.dup
-        }
```
beauby: In order for Hanami to work with Hanami::Model, the [Hanami loader](https://github.com/hanami/hanami/blob/master/lib/hanami/loader.rb#L134) will have to be updated to call `Hanami::Model::load!` instead of `Hanami::Model::duplicate`.

beauby:
Is this change in the [https://github.com/hanami/model/blob/new-engine/lib/hanami/model/configurator.rb#L38](configuration API) accidental? Before, the `adapter` method took a hash (`adapter type: :sql, uri: 'foo://bar'`) – or at least this is what the comments in `lib/bookshelf.rb` in a Hanami 0.8 generated app implies – and now it expects the type and uri to be given directly as parameters.

beauby:
Also, when one generates a model before the backing sql table is created, the whole thing breaks because schema inferring fails.

.travis.yml
beauby: IIRC consensus finally shifted to using HoundCI – should this be reverted?

README.md
```diff
@@ -256,253 +220,32 @@ This is a **huge improvement**, because:
...
-```
-
-### Data Mapper
-
-A persistence mapper that keeps entities independent from database details.
-It is database independent, it can work with SQL, document, and even with key/value stores.
-
-The role of a data mapper is to translate database columns into the corresponding attribute of an entity.
+However, there could be cases where there is a mismatch between the columns and attributes names.
+This is the case of **legacy databases**.
```
beauby: Suggestion (wording): 
> However, there are cases where columns and attribute names do not match (mainly **legacy databases**).

jodosha: @beauby Thanks for the suggestion: https://github.com/hanami/model/pull/334/commits/cb59e7e6d8e15f56cac203e2d1948aa686d465d0

.travis.yml
jodosha: @beauby Not yet. I just activated HoundCI for this project. From the next PR we can revert this diff change.

.travis.yml
jodosha: @cllns I prefer to avoid beta Travis CI features for now. Too moving parts.

jodosha:
Okay, I tried HoundCI and it doesn't respect our `.rubocop.yml`. My apologize for the the tons of emails you have received.

beauby:
@jodosha Looks like you forgot to add ` .hound.yml` with
```
ruby:
  config_file: .rubocop.yml
```

jodosha:
@beauby The first batch of comments was caused by the lack of `.hound.yml`, but then I added it: https://github.com/hanami/model/pull/334/commits/4ad9db6de4b97df13e9e7486b8c6c5d3ff84c2c9 and the second wave of comments happened.

Am I missing something?

beauby:
@jodosha You might want to add the following to the `.rubocop.yml`
```
Style/FrozenStringLiteralComment:
  Enabled: false
```

jodosha:
@beauby Sure, but why it doesn't report the issue locally and on Travis CI?

beauby:
@jodosha According to [github](https://github.com/bbatsov/rubocop/commits/ec3123fc3454b080e1100e35480c6466d1240fff/lib/rubocop/cop/style/frozen_string_literal_comment.rb), this cop only exists since early 2016, so I'd put my money on HoundCI running a fresher version of rubocop.

beauby:
Looking at [rubocop's default config](https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L521), it appears that by default, this cop is enabled or disabled depending on the TargetRubyVersion of rubocop which may be influenced by the environment in which it is ran, as it is not explicitly specified in the `.rubocop.yml`.
According to [this discussion](https://github.com/bbatsov/rubocop/issues/2739), the default value for this cop will be `disabled` anyways.

jodosha:
@beauby Thanks for the quick and detailed reply. I'm running rubocop 0.45.0 on MRI 2.3.1 and I didn't ran into that frozen string violation by running `bundle exec rubocop`.

---

My point is: local machine and TravisCI agree, while HoundCI doesn't. The first two environments are more transparent for us than the latter. What I want to avoid is to have headaches like this in the future.

beauby:
@jodosha Agreed, but TBH, this seems like a shortcoming on rubocop's side, as the environment should not have any effect on static analysis of code.

README.md
jodosha: Fixed, thanks!

hanami-model.gemspec
jodosha: Yes, thanks!

lib/hanami/model/mapping.rb
jodosha: The `associations` block is passed as it is to ROM. At this point we're just interested to capture `has_many` calls and discard the rest.

lib/hanami/model.rb
jodosha: @beauby `Hanami::Loader` doesn't exist anymore. We replaced it in Hanami master.

beauby: :+1:

jodosha:
This was merged in master. Thank you all! 🎉

PR #338, opened 2016-10-29
Fix Readme for repository usage

aderyabin:
A repository should not be instantiated. That is type in the README.


AlfonsoUceda:
@aderyabin I'm going to close this PR because repositories must be instantiated.

PR #340, opened 2016-11-11
Fix return value of Hanami::Repository#all.

beauby:
`Hanami::Repository#all` currently returns an instance of `ROM::Repository::RelationProxy`, which is inconsistent with the comment above the definition (`@return [Array<Hanami::Entity>] all the entities`), and the fact that all other query methods return materialized results.

This PR fixes this by calling `to_a` on the `RelationProxy`.

### DIFF
--- a/lib/hanami/repository.rb
+++ b/lib/hanami/repository.rb
@@ -384,7 +384,7 @@ module Hanami
     # @example
     #   UserRepository.new.all
     def all
-      root.as(:entity)
+      root.as(:entity).to_a
     end
 
     # Returns the first record for the relation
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
@beauby we need tests to avoid a regression, you should check the following tests and check all is an array.

https://github.com/hanami/model/blob/935d997eba3b452682a462ae8c7541806c25e11a/test/integration/repository/base_test.rb#L23-L30
https://github.com/hanami/model/blob/935d997eba3b452682a462ae8c7541806c25e11a/test/integration/repository/legacy_test.rb#L21-L28

beauby:
@AlfonsoUceda Done :+1:

AlfonsoUceda:
@beauby perfect, I think this PR will be part of v0.8.0, cc @jodosha

jodosha:
@beauby Thank you. I just merged it.

PR #341, opened 2016-11-13
Fix example in readme.

beauby:
Example was declaring

```
```

and was subsequently calling `UserRepository.new.update` instead of `repository.update`.


jodosha:
@beauby You're right. Thanks for catching this. 👍

PR #354, opened 2016-12-02
Introduce Repository#connection

### LINKED ISSUE #345: Perform raw sql in a Hanami::Repository subclass
I our app (pre 0.9) we use raw sql statements quite often to use our database to the fullest. To be able to upgrade to version 0.9 we would really need to have something similar like the fetch/execute methods that used to be available.
Can this be done right now?

### LINKED ISSUE #351: Count method
I think it will be awesome to create `COUNT` command for repository object. What do you think?
If you agree I can send PR 😊

jodosha:
This is a proposal to introduce a new private method for repositories: `#connection`.

It returns a low level connection for the current database. This allows developers to perform low level operations like fetching raw data with hand written SQL queries, or to perform a SQL `COUNT`.

---

We want that the public interface of `Repository` to be smallest as possible in order to make it compatible with all the available adapters. This is the reason why [we rejected the idea](https://github.com/hanami/model/issues/351#issuecomment-264388994) of having `#count`.

But at the same time, we want to allow developers to perform low level operations that they may need.

**NOTE:** Using `#connection` may limit the portability of your repository. For instance, if you implement a method that fetches data via a SQL query, then it would be hard to swap the adapter to Redis, unless you change that fetching logic.

---

### Fetch raw data via SQL

```ruby
  def find_active_users
    connection.fetch("SELECT * FROM users WHERE active='t'").to_a
  end
end

data = repository.find_active_users

data.class # => Array
data.first # => {:id=>1, :name=>"Luca", :age=>34}
```

**NOTE:** Please note that `connection#fetch` is only available for the current adapter, and it may change if you use a different adapter. Again, using `#connection` limits the portability, you're warned 😉 

In the past we used to have `Repository#fetch`, and I think this is wrong for encapsulation. This encourages developers to leak raw fetching scripts (eg a SQL query) **outside** of the repository.

```ruby
repository.fetch("SELECT * FROM users WHERE active='t'").to_a # WRONG
```

For the same reason we do **NOT** want `repository.where(active: true)` is possible, we should forbid `#fetch`.

Even if `#connection` limits portability, at least developers are forced to **expose explicit, meaningful methods**. In the example above, we defined `#find_active_users`, which abstracts that logic on the outside of the repository.

### Execute raw SQL commands

```ruby
  def reset_comments_count
    connection.execute("UPDATE users SET comments_count = 0")
  end
end
```

This is similar to the previous example. Again, it limits portability because `connection#execute` may not be available for other adapters, but it has the advantage of forcing devs to expose meaningful methods

### Count records

```ruby
  def count
    users.count
  end

  def active_users
    users.where(active: true)
  end

  def count_active_users
    active_users.count
  end

  def count_users_via_connection # silly name, I know ;-)
    connection[:users].count
  end
end
```

The method `connection#[]` is derived from Sequel (`sequel` gem).

---

Closes #345 
Closes #351 

---

/cc @hanami/core-team @hanami/contributors @choallin @solnic @flash-gordon for review.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,3 +25,7 @@ Style/GuardClause:
 Metrics/BlockLength:
   Exclude:
     - "test/**/*.rb" # Minitest syntax generates this false positive
+Bundler/OrderedGems:
+  Enabled: false
+Style/EmptyMethod:
+  Enabled: false
--- a/lib/hanami/model.rb
+++ b/lib/hanami/model.rb
@@ -17,7 +17,7 @@ module Hanami
 
     # @api private
     # @since 0.7.0
-    @__repositories__ = Concurrent::Array.new # rubocop:disable Style/VariableNumber
+    @__repositories__ = Concurrent::Array.new
 
     class << self
       # @since 0.7.0
--- a/lib/hanami/model/error.rb
+++ b/lib/hanami/model/error.rb
@@ -8,7 +8,7 @@ module Hanami
     class Error < ::StandardError
       # @api private
       # @since 0.7.0
-      @__mapping__ = Concurrent::Map.new # rubocop:disable Style/VariableNumber
+      @__mapping__ = Concurrent::Map.new
 
       # @api private
       # @since 0.7.0
--- a/lib/hanami/repository.rb
+++ b/lib/hanami/repository.rb
@@ -423,6 +423,69 @@ module Hanami
 
     private
 
+    # Gives access to low level connection in order to perform specific
+    # operations for the current database.
+    #
+    # NOTE: Please note that using `#connection` will limit the portability of
+    # the repository in case you want to switch the database/adapter.
+    # For instance, if you fetch data with a raw SQL query, then it would be
+    # impossible to use a Redis database without changing that query.
+    #
+    # NOTE: In order to use `#connection`, please read the API docs of your
+    # current adapter.
+    # In case you're using a SQL database, please check Sequel (`sequel` gem)
+    # API docs.
+    #
+    # @return [Object] A raw connection, the type depends on the current adapter
+    #
+    # @since x.x.x
+    #
+    # @example Fetch data with raw SQL
+    #   class UserRepository < Hanami::Repository
+    #     def find_all
+    #       connection.fetch('SELECT * FROM users').to_a
+    #     end
+    #   end
+    #
+    #   repository = UserRepository.new
+    #   users      = repository.find_all
+    #
+    #   users.class # => Array
+    #   users.first # => {:id=>1, :name=>"Luca", :age=>34, :created_at=>2016-12-02 08:05:56 UTC, :updated_at=>2016-12-02 08:05:56 UTC}
+    #
+    #   # NOTE that `#fetch` is only available for SQL adapter
+    #
+    # @example Count records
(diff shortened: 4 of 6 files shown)


solnic:
What if we add `ROM::SQL::Relation#read(sql_str)` that returns a relation? Then you could do `def find_active_users; users.read("SELECT ...").to_a; end`. Key benefit is that you keep working with relations this way and you can use all their extra features, I actually just implemented it, see:

``` ruby
irb(main):014:0> user_repo.users.read('select id, name from users').as(:entity).one
=> #<User:0x007fbd39d36e88 @attributes={:id=>1, :name=>"Piotr"}>
```

choallin:
It for sure will work for us.  
We did use the old Repository#fetch as private method so, it will not be this much of work to do to upgrade since the connection#fetch will be private as well.
I really like the idea that it would be possible to use the relationships like @solnic mentioned *together* with the option to just get the raw dataset. This is necessary for us, because we have a repository that connects to a oracle database and fetches from various tables, so we don't have an entity for them.

solnic:
> I really like the idea that it would be possible to use the relationships like @solnic mentioned together with the option to just get the raw dataset. This is necessary for us, because we have a repository that connects to a oracle database and fetches from various tables, so we don't have an entity for them

Are you working with plain hashes or do you transform them in some way? If we add a `read` method like I showed in my prev comment, it will be possible to define custom mappers for your relations. This is very powerful as you can use rom mappers for that, and they will be automatically available for each relation. In example you'd be able to do `user_repo.users.read("...").as(:my_custom_mapper)`. In the early days of rom we didn't have auto-mapping, so we started with custom mappers, they are still used by repositories generate them for you **but** in case of custom (and potentially complex) queries, you can still use your own mappers. This allows you to do things like expanding a big result set into nested structures, return custom objects from hashes, perform value coercions and so on. I consider this as pretty advanced usage, but there are cases where people may need it. Just mentioning that it's totally possible and works OOTB with relations.

choallin:
In the use case I described above we are working with plain hashes. They are just data to be displayed in a table. Nothing interesting about them... .
But in other cases a mapping to an entity would be great. The way you described it is especially useful because we can load all dependencies for a business object and create all related entities at once. 

Just a side node: The read method still is a private method in the Repositories, right?

solnic:
> Just a side node: The read method still is a private method in the Repositories, right?

It'd be a relation method, so doing `user_repo.users.read` in app's code would be a smell, it would be a private method in terms of its visibility. We could also have `Hanami::Repository#read` which would delegate to its relation and make that private.

jodosha:
> We could also have Hanami::Repository#read which would delegate to its relation and make that private.

I'd not expose it directly, otherwise people can abuse via `repo.read("SELECT ...")`. Even the other proposals above `repo.users.read("...")` don't make it really private, but at least we're making harder to do the wrong thing.

solnic:
> I'd not expose it directly

That's why I said it could be a private repository method :)

AMHOL:
I also like that `#read` encourages `CQRS` simply by expressing intent with the method name, for example `connection.execute("UPDATE users SET comments_count = 0")` should probably be handled by a command

TiteiKo:
👍  to me too

jodosha:
Closing in favor of https://github.com/hanami/model/pull/362

PR #357, opened 2016-12-11
Timestamp plugin respects given timestamps.

thhermansen:
If no timestamps is given, as it most often will, then current time
will be used. If timestamps are given, the given timestamps will be
used.

This restores pre-0.7 behaviour and I think it is correct that when you
ask the repository to insert/update a row with given data it shouldn’t
change given data before data is persisted to the database.

### DIFF
--- a/lib/hanami/model/plugins/timestamps.rb
+++ b/lib/hanami/model/plugins/timestamps.rb
@@ -63,7 +63,7 @@ module Hanami
           # @since 0.7.0
           # @api private
           def _touch(value, now)
-            value[:updated_at] = now
+            value[:updated_at] ||= now
             value
           end
         end
@@ -79,7 +79,7 @@ module Hanami
           # @api private
           def _touch(value, now)
             super
-            value[:created_at] = now
+            value[:created_at] ||= now
             value
           end
         end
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@thhermansen thanks for this PR, what is the case you need to add timestamps yourself? It makes sense respect your passed timestamps but I'd like to know your use case.

Please, can you remove the commit for rubocop? don't worry about that, we need to add two new rules to avoid that problem, sorry for the problems.

thhermansen:
Removed the rubocop commit now.

I had some repository tests where I inserted rows with different `created_at`. It allowed me to test a repository method which is filtering based on `created_at`. So it is a bit obscure use case, but I can see it being useful in some situation to dictate the timestamps for records as well outside of tests.

The current implementation in master also allows for `created_at` being set on `repository.update`, but not on create. `updated_at` is never allowed to be set. I find it better to allow these attributes to be set every time.

Thanks for the work on Hanami Model and migration to ROM btw! :-)

AlfonsoUceda:
@jodosha what do you think about this? To me it's ok, it will help to create better tests as @thhermansen said.

jodosha:
@thhermansen Thank you. Merging this 👍

PR #358, opened 2016-12-11
Allow Hanami::Entity to be used in Schema as type.

thhermansen:
This allows for Hanami::Entity being used in schema like:

```ruby
class Account < Hanami::User
end

  attributes do
    attribute :owner, Types::Entity(User)
  end
end
```

`Types::Entity(User)` is new, it matches the `Types::Collection(User)`. Another solution would be to make `User` entity class compatible with dry types.

-----------------

I removed the part of the code which symbolizes both attributes for schemaless entities and entites with schema. Calling `symbolize!` on attributes returned by schema will call `to_hash` on the `User` which defeats the purpose of expressing that `owner` should be a User object. I'm not sure if I'm 100% happy with this change. The previous implementation made it clear when hash was symbolised, but I don't know if this is possible to do this simpler as long as Entity responds to `to_hash`. Keeping entities respond to `to_hash` makes them compatible to be used as data and passed to `repository#create/update` which I like.

I have no idea if this is the best fix and I guess you may want to solve this differently as my knowledge to Hanami Model as a whole is limited, so take this pull request more like a suggestion :-) I also suspect that this touches aspects which also must work when belongs to association is completed.

If this is merged it will close https://github.com/hanami/model/issues/356.

### DIFF
--- a/lib/hanami/entity/schema.rb
+++ b/lib/hanami/entity/schema.rb
@@ -74,11 +74,13 @@ module Hanami
         # @since 0.7.0
         # @api private
         def call(attributes)
-          if attributes.nil?
-            {}
-          else
-            attributes.dup
-          end
+          attributes = if attributes.nil?
+                         {}
+                       else
+                         attributes.dup
+                       end
+
+          Utils::Hash.new(attributes).symbolize!
         end
 
         # @since 0.7.0
@@ -205,9 +207,8 @@ module Hanami
       # @since 0.7.0
       # @api private
       def call(attributes)
-        Utils::Hash.new(
-          schema.call(attributes)
-        ).symbolize!
+        attributes = schema.call(attributes)
+        Utils::Hash.new(attributes)
       end
 
       # @since 0.7.0
--- a/lib/hanami/model/types.rb
+++ b/lib/hanami/model/types.rb
@@ -18,11 +18,15 @@ module Hanami
       #
       # @since 0.7.0
       module ClassMethods
+        def Entity(type) # rubocop:disable Style/MethodName
+          Schema::CoercibleType.wrap_unless_dry_type(type)
+        end
+
         # Define an array of given type
         #
         # @since 0.7.0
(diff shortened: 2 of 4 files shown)


Flink:
@thhermansen Are you sure this is working? I’m trying to replicate your PR in a project using Hanami 0.9.2 and it still returns a `Hanami::Utils::Hash` instead of the desired type 😕

Flink:
@thhermansen It seems to be me doing something wrong, don’t worry about my previous comment 😅

Flink:
After some more tests, it still doesn’t work for me (the code in `schema.rb` seems to have evolved since) 😕 I still have a `Hanami::Utils::Hash` object instead of the model one.

Flink:
From what I understand at the moment, the code seems to have evolved and even `Types::Collection` is not doing what it should. I tested a custom attribute with a collection of my model and it returns `ROM::Struct[Category]` instead of `Category`.

solnic:
@Flink could you provide a repro script?

Flink:
@solnic https://gist.github.com/Flink/cbcca82bda849b1ecfd22edd9d068bf8 With this, the `parents` attribute should return a `ROM::Struct[Category]`. I tried to do this since I was hoping to get a collection of `Category` objects and then just use the first one.

Flink:
@solnic For the `belongs_to` association to the same model, I found this to be working for returning `Category` objects (it’s an ugly hack 😬):
```ruby
class Category < Hanami::Entity
  # Ugly hack for now
  def initialize(attributes = nil)
    # don’t understand why `optional` is needed here but it won’t work without it because of `nil`s
    self.class.schema.send(:schema).type.options[:member_types][:parent] =
      self.class.schema.send(:schema).type.options[:member_types][:category].optional
    super
  end

  def parent
    return unless attributes[:parent]
    Category.new(attributes[:parent])
  end
end
```

jodosha:
https://github.com/hanami/model/issues/356#issuecomment-274113633

PR #359, opened 2016-12-13
Fix formatting text for migration error message

davydovanton:
### Before
screenshot 2016-12-14 02 03 16

### After
screenshot 2016-12-14 02 03 08


beauby:
IMO the right fix would be:
```ruby
       DB_CREATION_ERROR = "Database creation failed. If the database exists, " \
                             "then its console may be open. See this issue for more details: " \
                             "https://github.com/hanami/model/issues/250".freeze
```

davydovanton:
@beauby good catch, thanks 👍 
fixed

PR #362, opened 2016-12-23
Added specs for relation read

jodosha:
This is a new feature introduced by `rom-sql` `0.9.1`, as result of the following discussions:

  * https://github.com/hanami/model/issues/351
  * https://github.com/hanami/model/pull/354
  * https://github.com/hanami/model/pull/342
  * https://github.com/hanami/model/issues/345

---

This feature allows to read entities starting from a raw query string (eg SQL).

Because it's exposed by ROM relations, the purpose of this PR is to add a unit test for it to prevent regressions.


AlfonsoUceda:
@jodosha it's perfect, let me point out, this is the [commit](https://github.com/rom-rb/rom-sql/commit/aaf10ec4eaf40f578bf9b325fc02b62cd356d779) where @solnic added that feature.

PR #366, opened 2017-01-03
Map rom/rom-sql exceptions and re-raise as hanami-model errors

jodosha:
As we are approaching 1.0, we want to provide a stable set of exceptions that developers can rescue and handle.

For this purpose we need to map ROM exceptions and re-raise as Hanami::Model errors.

Ref: https://github.com/hanami/model/issues/361

### DIFF
--- a/lib/hanami/model/error.rb
+++ b/lib/hanami/model/error.rb
@@ -88,5 +88,14 @@ module Hanami
         super
       end
     end
+
+    # Unknown database type error for repository auto-mapping
+    #
+    # @since x.x.x
+    class UnknownDatabaseTypeError < Error
+      def initialize(message)
+        super
+      end
+    end
   end
 end
--- a/lib/hanami/model/error.rb
+++ b/lib/hanami/model/error.rb
@@ -93,9 +93,6 @@ module Hanami
     #
     # @since x.x.x
     class UnknownDatabaseTypeError < Error
-      def initialize(message)
-        super
-      end
     end
   end
 end
--- a/lib/hanami/model.rb
+++ b/lib/hanami/model.rb
@@ -78,9 +78,12 @@ module Hanami
       repositories.each(&:load!)
 
       @container = ROM.container(configuration)
+
       configuration.define_entities_mappings(@container, repositories)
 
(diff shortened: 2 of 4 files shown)


lib/hanami/model/error.rb
```diff
@@ -88,5 +88,14 @@ def initialize(message = 'Check constraint has been violated')
         super
       end
     end
+
+    # Unknown database type error for repository auto-mapping
+    #
+    # @since x.x.x
+    class UnknownDatabaseTypeError < Error
+      def initialize(message)
```
pascalbetz: `initialize` needed?

AlfonsoUceda: http://stackoverflow.com/a/3382882/1180580

AlfonsoUceda:
👍

lib/hanami/model/error.rb
pascalbetz: @AlfonsoUceda there is no default message...

AlfonsoUceda: Yes, I already see, maybe it's a typo.

solnic: @jodosha actually, in rom-sql 1.0.0 inferrer will no longer raise that error, so that you can define missing attributes yourself. Can we expose schema definition in hanami repos? (in a similar way as assoc, I suppose). The reason for this change is flexibility, because we can't guarantee that inferrer will always work and that even if it works, the attributes are defined exactly like you want. Having a way of combining inference with explicit definitions is a win-win in such case, because you avoid boilerplate and still have complete control.

jodosha: @solnic Thanks for let us to know. So this PR is not needed, right?

jodosha: @pascalbetz @AlfonsoUceda Yes, it's a leftover. Thanks for spotting it! 😄

jodosha: UPDATE: I'm still shipping this PR, because we still want to increase the coverage of ROM exceptions. If `ROM.container` raises another error we'll be able to map it.

PR #367, opened 2017-01-03
Introducing Configuration callbacks

### LINKED ISSUE #352: Customisations of Sequel
Sometimes there is a need to access lower level details of hanami-model to configure some use case specific stuff.

For example, if there are infinite timestamps in DB, Sequel will throw error. With access to the Sequel connection this could be easily solved:
`connection.convert_infinite_timestamps = 'nil'`

There are some other possible cases when this would be helpful. Such as DB sharding, transactions, turning on specific Sequel plugins, etc

It could be useful to have access to configuration details, which are hidden right now

@jodosha ticket as result of gitter chat discussion

### LINKED ISSUE #361: ROM::SQL::UnknownDBTypeError exception is raised when a PostgreSQL table has an enum column
ROM raises `ROM::SQL::UnknownDBTypeError` when it attempts to infer the type for an enum column, which prevents the application from being loaded.

Steps to reproduce:

% hanami new bookshelf --database=postgres
% cd bookshelf
% bundle install
% bundle exec hanami generate migration create_books
```ruby
Hanami::Model.migration do
  change do
    extension :pg_enum
    create_enum :form, %w(novel poem drama short_story novella)

    create_table :books do
      primary_key :id

      column :title,  String, null: false
      column :author, String, null: false
      column :form, :form, null: false

      column :created_at, DateTime, null: false
      column :updated_at, DateTime, null: false
    end
  end
end
```
% bundle exec hanami db prepare
% bundle exec hanami generate model book
% bundle exec hanami console
```
bundler: failed to load command: hanami (/usr/local/var/rbenv/versions/2.3.3/bin/hanami)
ROM::SQL::UnknownDBTypeError: Cannot find corresponding type for form
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:69:in `block in map_type'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:68:in `fetch'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:68:in `map_type'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/extensions/postgres/inferrer.rb:48:in `map_type'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:56:in `build_type'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:45:in `block in call'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:44:in `each'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:44:in `each_with_object'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema/inferrer.rb:44:in `call'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/relation.rb:47:in `block (2 levels) in inherited'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/schema.rb:83:in `finalize!'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-sql-0.9.0/lib/rom/sql/schema.rb:24:in `finalize!'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize/finalize_relations.rb:53:in `build_relation'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize/finalize_relations.rb:25:in `block (2 levels) in run!'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize/finalize_relations.rb:24:in `each'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize/finalize_relations.rb:24:in `block in run!'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/relation_registry.rb:6:in `initialize'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize/finalize_relations.rb:23:in `new'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize/finalize_relations.rb:23:in `run!'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize.rb:93:in `load_relations'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/setup/finalize.rb:62:in `run!'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/create_container.rb:35:in `finalize'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/create_container.rb:13:in `initialize'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/create_container.rb:54:in `initialize'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/create_container.rb:59:in `new'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rom-2.0.2/lib/rom/create_container.rb:59:in `container'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-model-0.7.0/lib/hanami/model.rb:80:in `load!'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components/components.rb:57:in `block (2 levels) in <module:Components>'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components/component.rb:42:in `call'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:85:in `block (2 levels) in resolve'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/concurrent-ruby-1.0.3/lib/concurrent/map.rb:133:in `block in fetch_or_store'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/concurrent-ruby-1.0.3/lib/concurrent/map.rb:122:in `fetch'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/concurrent-ruby-1.0.3/lib/concurrent/map.rb:132:in `fetch_or_store'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:83:in `block in resolve'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:82:in `each'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:82:in `resolve'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components/component.rb:136:in `resolve_requirements'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components/component.rb:34:in `call'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:85:in `block (2 levels) in resolve'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/concurrent-ruby-1.0.3/lib/concurrent/map.rb:133:in `block in fetch_or_store'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/concurrent-ruby-1.0.3/lib/concurrent/map.rb:122:in `fetch'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/concurrent-ruby-1.0.3/lib/concurrent/map.rb:132:in `fetch_or_store'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:83:in `block in resolve'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:82:in `each'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/components.rb:82:in `resolve'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/commands/command.rb:47:in `initialize'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/commands/console.rb:47:in `initialize'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/cli.rb:84:in `new'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/cli.rb:84:in `console'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/invocation.rb:126:in `invoke_command'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/base.rb:444:in `start'
  /usr/local/var/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/bin/hanami:5:in `<top (required)>'
  /usr/local/var/rbenv/versions/2.3.3/bin/hanami:22:in `load'
  /usr/local/var/rbenv/versions/2.3.3/bin/hanami:22:in `<top (required)>'
```

I tried defining the enum type in the entity's [custom schema](http://hanamirb.org/guides/models/entities/#custom-schema), but it didn't work.
```ruby
class Book < Hanami::Entity
  attributes do
    attribute :id,         Types::Int
    attribute :title,      Types::String
    attribute :author,     Types::String
    attribute :form,       Types::Strict::String.enum('novel', 'poem', 'drama', 'short_story', 'novella')
    attribute :created_at, Types::Time
    attribute :updated_at, Types::Time
  end
end
```

jodosha:
The goal is to provide access to the raw connection during the configuration setup.

In case of a SQL database, it returns a raw Sequel connection: eg. `Sequel::Postgres::Database`. It can be used to setup it (see #352) or to activate extensions.

Example:

```ruby
Hanami::Model.configure do
  adapter :sql, ENV['DATABASE_URL']

  before do |connection|
    connection.extension :pg_enum
  end
end
```

That `before` block will be yielded as first step during `Hanami::Model.load!`.

---

@hanami/core @hanami/issues @hanami/ecosystem @vladra @Moratorius @solnic @flash-gordon @AMHOL I'd love to hear your feedback.

---

Closes #352 
Closes #361

### DIFF
--- a/lib/hanami/model/configuration.rb
+++ b/lib/hanami/model/configuration.rb
@@ -23,6 +23,7 @@ module Hanami
         super(configurator.backend, configurator.url)
         @migrations   = configurator._migrations
         @schema       = configurator._schema
+        @callbacks    = configurator.callbacks
         @mappings     = {}
         @entities     = {}
       end
@@ -96,6 +97,14 @@ module Hanami
           entity   = r.entity
 
           entity.schema = Sql::Entity::Schema.new(entities, container.relations[relation], mappings.fetch(relation))
+        end
+      end
+
+      # @since x.x.x
+      # @api private
+      def run_before_callbacks
+        @callbacks[:before].each do |cb|
+          cb.call(connection)
         end
       end
     end
--- a/lib/hanami/model/configurator.rb
+++ b/lib/hanami/model/configurator.rb
@@ -25,10 +25,20 @@ module Hanami
       # @api private
       attr_reader :_schema
 
+      # @since x.x.x
+      # @api private
+      attr_reader :callbacks
+
       # @since 0.7.0
       # @api private
       def self.build(&block)
         new.tap { |config| config.instance_eval(&block) }
+      end
+
+      # @since x.x.x
+      # @api private
+      def initialize
+        @callbacks = Hash[before: [], after: []]
       end
 
       private
@@ -56,6 +66,12 @@ module Hanami
       # @api private
       def schema(path)
         @_schema = path
+      end
+
+      # @since x.x.x
+      # @api private
+      def before(&blk)
+        callbacks[:before] << blk
       end
     end
   end
--- a/lib/hanami/model.rb
+++ b/lib/hanami/model.rb
@@ -73,6 +73,7 @@ module Hanami
 
     # @since 0.1.0
     def self.load!(&blk) # rubocop:disable Metrics/AbcSize
+      configuration.run_before_callbacks
       configuration.setup.auto_registration(config.directory.to_s) unless config.directory.nil?
       configuration.instance_eval(&blk)                            if     block_given?
       repositories.each(&:load!)
--- a/test/fixtures/database_migrations/20170103142428_create_colors.rb
+++ b/test/fixtures/database_migrations/20170103142428_create_colors.rb
@@ -1,0 +1,22 @@
+Hanami::Model.migration do
+  change do
+    case Database.engine
+    when :postgresql
+      extension :pg_enum
+      create_enum :rainbow, %w(red orange yellow green blue indigo violet)
+
+      create_table :colors do
+        primary_key :id
+
+        column :name, :rainbow, null: false
+
+        column :created_at, DateTime, null: false
+        column :updated_at, DateTime, null: false
+      end
+    else
+      create_table :colors do
+        primary_key :id
+      end
+    end
+  end
+end
--- a/test/integration/repository/base_test.rb
+++ b/test/integration/repository/base_test.rb
@@ -534,6 +534,23 @@ describe 'Repository (base)' do
           found.must_equal(product)
         end
       end
+
+      describe 'enum database type' do
+        it 'allows to write data' do
+          repository = ColorRepository.new
+          color = repository.create(name: 'red')
+
+          color.must_be_kind_of(Color)
+          color.name.must_equal 'red'
+        end
+
+        it 'raises error if the value is not included in the enum' do
+          repository = ColorRepository.new
+
+          exception = -> { repository.create(name: 'grey') }.must_raise(Hanami::Model::Error)
+          exception.message.must_equal %("grey" (String) has invalid type for :name)
+        end
+      end
     end
   end
 end
--- a/test/support/database.rb
+++ b/test/support/database.rb
@@ -42,6 +42,10 @@ module Database
   def self.engine
     ENV['HANAMI_DATABASE_TYPE'].to_sym
   end
+
+  def self.engine?(name)
+    engine == name.to_sym
+  end
 end
 
 Database::Setup.new.run
--- a/test/support/database/strategies/sql.rb
+++ b/test/support/database/strategies/sql.rb
@@ -20,6 +20,10 @@ module Database
           adapter    ENV['HANAMI_DATABASE_ADAPTER'].to_sym, ENV['HANAMI_DATABASE_URL']
           migrations Dir.pwd + '/test/fixtures/database_migrations'
           schema     Dir.pwd + '/tmp/schema.sql'
+
+          before do |connection|
+            connection.extension(:pg_enum) if Database.engine?(:postgresql)
+          end
         end
       end
 
--- a/test/support/fixtures.rb
+++ b/test/support/fixtures.rb
@@ -38,6 +38,9 @@ end
 class Product < Hanami::Entity
 end
 
+class Color < Hanami::Entity
+end
+
 class UserRepository < Hanami::Repository
   def by_name(name)
     users.where(name: name).as(:entity)
@@ -129,6 +132,9 @@ class WharehouseRepository < Hanami::Repository
 end
 
 class ProductRepository < Hanami::Repository
+end
+
+class ColorRepository < Hanami::Repository
 end
 
 Hanami::Model.load!


lib/hanami/model/configurator.rb
```diff
@@ -25,12 +25,22 @@ class Configurator
...
       # @since 0.7.0
       # @api private
       def self.build(&block)
         new.tap { |config| config.instance_eval(&block) }
       end
 
+      # @since x.x.x
+      # @api private
+      def initialize
+        @callbacks = Hash[before: [], after: []]
```
pascalbetz: What are `:after` callbacks for?

AlfonsoUceda:
:+1: just to say the same of @pascalbetz

lib/hanami/model/configuration.rb
```diff
@@ -98,6 +99,14 @@ def define_entities_mappings(container, repositories)
           entity.schema = Sql::Entity::Schema.new(entities, container.relations[relation], mappings.fetch(relation))
         end
       end
+
+      # @since x.x.x
+      # @api private
+      def run_before_callbacks
+        @callbacks[:before].each do |cb|
```
Moratorius: My only reservation is that `before` callbacks are limited to configuring the connection.  It may be more flexible to yield an object which has the connection as an accessor, but if you don't anticipate needing that functionality, then this is fine.

solnic:
Maybe it's a premature concern, but actually `Gateway#connection` is an adapter-specific thing. We have rom adapters without any connection object.

lib/hanami/model/configuration.rb
jodosha: @Moratorius Which kind of object?

jodosha:
@solnic Thanks for pointing that. Which adapters? HTTP?

lib/hanami/model/configurator.rb
jodosha: I forgot to mention this in the description. Do you think we need them too?

solnic:
> @solnic Thanks for pointing that. Which adapters? HTTP?

any adapter without a database, so rom-http rom-yaml rom-csv rom-json rom-git and whatever we will come up in the future. In fact, even if there's a database we may have cases where there's no connection object for whatever reason.

I think the appropriate thing to do is to yield a gateway to this block, because you will always have gateways. Also, I should mention that rom-sql supports `:extensions` option which can be set to an array with additional extensions that should be loaded by sequel, but in hanami-model you only support uri string at the moment.

jodosha:
@solnic Thanks for letting us to know about `:extensions`. I wanted to introduce this yield block because sometimes users [need to set values on the connection](https://github.com/hanami/model/issues/352). Eg

```ruby
connection.convert_infinite_timestamps = nil
```

solnic:
@jodosha OK so `gateway.connection.convert_inifite_timestamps = nil` but eventually we just need an interface for that on gateways, I don't want sequel connection to leak outside of the gateway tbh

jodosha:
@solnic I also have another question. I imagine that the connection may change during the lifetime of a Ruby process. Is that correct?

solnic:
@jodosha yeah there is a risk, but it's not intentional from rom's pov, it's just how some drivers are implemented

lib/hanami/model/configurator.rb
pascalbetz: Hmm. Not sure `:before` and `:after` are even the right names..._before_ what? _after_ what? 
It's not that this is executed everytime before the connection is used. It's more a hook to _setup_ the connection?

mereghost: Maybe we could add them if needed later on, unless someone comes with a use case.

For the names I do agree with @pascalbetz , before and after are very generic but not that much descriptive.

jodosha: @pascalbetz @mereghost Yes, I agree with you. But we aim for actionable feedbacks. :wink:  Share your crazy names for this stuff! :joy_cat:

jodosha:
Folks, an alternative design could be:

```ruby
Hanami::Model.configure do
  adapter :sql, ENV['DATABASE_URL']

  gateway do |g|
    g.connection.convert_infinite_timestamps = nil
    g.connection.extension :pg_enum
  end
end
```

This solves the naming problems (`before`/`after`, wuh?), and yields a gateway object. We can warn that `connection` isn't always available for all the adapters.

@pascalbetz @mereghost @solnic @AlfonsoUceda WDYT?

lib/hanami/model/configurator.rb
pascalbetz: well...if only i had a crazy idea:-)
I like the `gateway` idea you presented. Not sure if `gateway` has limitations for the future? But if I understood the comment from @solnic there will always be a `gateway` but not necessarily a `connection` so 'gateway'
 is even better.

Moratorius:
I like the `gateway` dsl; it's easier to infer its purpose instead of a generic `before` hook that happens to yield the connection.

solnic:
For extensions it's only a matter of adding support for connection options, so that hanami passes both URI and options to the gateway constructor. Additional settings could be configured via something like `Gateway#set(some_opts)` and it would just go through opts and do `opts.each { |k, v| connection.send("#{k}=", v) }`. WDYT? Then you could do this:

``` ruby
gateway do |g|
  g.set(convert_infinite_timestamps = nil)
end
```

mereghost:
@solnic If you intend to `rom` not leak the connection, your proposal would make it future-proof, but would demand some heavy documentation explaining what it does and how it works and where to look for available settings.

I'm not against it, just thinking about how magical it would seem and how we can explain it to newcomers.

jodosha:
@solnic Is that `g.set` possible as now, or is something that we should add in ROM? If so, is it compatible with your release plans?

solnic:
> @solnic Is that g.set possible as now, or is something that we should add in ROM? If so, is it compatible with your release plans?

It's not, that was just my idea how this could be done

jodosha:
Closing in favor of https://github.com/hanami/model/pull/372

PR #368, opened 2017-01-03
Logging

### LINKED ISSUE #277: Cannot log raw SQL
Hi.

I tried finding a way to log the requests being sent to the server but I couldn't find anything related to that in the code. Is there a way to do this that I'm unaware of or this is not currently possible? I saw Sequel has an option but that would mean having direct access to the DB object and I see it's under a couple of layers in the code and I couldn't find a way to reach it simply (which of course seems logical by design).

Thanks

jodosha:
This is a proposal to add:

  * Optional logging for database operations while the application is running
  * Automatic logging to `$stdout` when running migrations

---

## Logging for database operations

This is **optional** when using `hanami-model` in **standalone**, but it will be **enabled by default** for Hanami projects (`hanami` gem).

### Usage

```ruby
Hanami::Model.configure do
  # ...
  logger "log/development.log", level: :debug
end
```

Where the first argument is a `IO` Ruby object: a path to a file, or `$stdout`.

Again, this will be setup **automatically** in Hanami projects

### Example

```
app=hanami.model severity=INFO time=2017-01-03 17:17:17 UTC message=(0.000260s) SELECT `id`, `name`, `created_at`, `updated_at` FROM `authors` WHERE (`id` IN (1)) ORDER BY `authors`.`id`
```

## Automatic Logging for Migrations

This is **always enabled** without any configuration.

### Example

```
[migration] [INFO] Begin applying migration 20160831090612_add_rating_to_reviews.rb, direction: up
[migration] [INFO] (0.000209s) BEGIN
[migration] [INFO] (0.004497s) ALTER TABLE "reviews" ADD COLUMN "rating" integer DEFAULT 0
[migration] [INFO] (0.000191s) INSERT INTO "schema_migrations" ("filename") VALUES ('20160831090612_add_rating_to_reviews.rb') RETURNING "filename"
[migration] [INFO] (0.001470s) COMMIT
[migration] [INFO] Finished applying migration 20160831090612_add_rating_to_reviews.rb, direction: up, took 0.007864 seconds
```

This is too much verbose IMO, but we don't have control on it, as we use `Sequel` logger for that.

---

/cc @hanami/issues @hanami/ecosystem @davidpelaez for review. :+1: 

---

Closes #277

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,25 @@ puts user.created_at.to_s # => "2016-09-19 13:40:13 UTC"
 puts user.updated_at.to_s # => "2016-09-19 13:40:16 UTC"
 ```
 
+## Configuration
+
+### Logging
+
+In order to log database operations, you can configure a logger:
+
+```ruby
+Hanami::Model.configure do
+  # ...
+  logger "log/development.log", level: :debug
+end
+```
+
+It accepts the following arguments:
+
+  * `stream`: a Ruby StringIO object - it can be `$stdout` or a path to file (eg. `"log/development.log"`) - Defaults to `$stdout`
+  * `:level`: logging level - it can be: `:debug`, `:info`, `:warn`, `:info`, `:warn`, `:error`, `:fatal`, `:unknown` - Defaults to `:debug`
+  * `:formatter`: logging formatter - it can be: `:default` or `:json` - Defaults to `:default`
+
 ## Versioning
 
 __Hanami::Model__ uses [Semantic Versioning 2.0.0](http://semver.org)
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ end
 It accepts the following arguments:
 
   * `stream`: a Ruby StringIO object - it can be `$stdout` or a path to file (eg. `"log/development.log"`) - Defaults to `$stdout`
-  * `:level`: logging level - it can be: `:debug`, `:info`, `:warn`, `:info`, `:warn`, `:error`, `:fatal`, `:unknown` - Defaults to `:debug`
+  * `:level`: logging level - it can be: `:debug`, `:info`, `:warn`, `:error`, `:fatal`, `:unknown` - Defaults to `:debug`
   * `:formatter`: logging formatter - it can be: `:default` or `:json` - Defaults to `:default`
 
 ## Versioning
--- a/lib/hanami/model.rb
+++ b/lib/hanami/model.rb
@@ -73,6 +73,7 @@ module Hanami
 
     # @since 0.1.0
     def self.load!(&blk) # rubocop:disable Metrics/AbcSize
+      configuration.gateway.use_logger(configuration.logger)       unless configuration.logger.nil?
       configuration.setup.auto_registration(config.directory.to_s) unless config.directory.nil?
       configuration.instance_eval(&blk)                            if     block_given?
       repositories.each(&:load!)
--- a/lib/hanami/model/migrator.rb
+++ b/lib/hanami/model/migrator.rb
@@ -237,9 +237,9 @@ module Hanami
       #
       # @since 0.7.0
       # @api private
-      def initialize(configuration: self.class.configuration, stream: $stdout)
+      def initialize(configuration: self.class.configuration)
         @configuration = configuration
-        @adapter       = Adapter.for(configuration, stream)
+        @adapter       = Adapter.for(configuration)
       end
 
       # @since 0.7.0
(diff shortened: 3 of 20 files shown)


lib/hanami/model/migrator.rb
```diff
@@ -237,9 +237,9 @@ def self.version
       #
       # @since 0.7.0
       # @api private
-      def initialize(configuration: self.class.configuration)
+      def initialize(configuration: self.class.configuration, stream: $stdout)
```
pascalbetz: Why passing the `stream` and not a `Logger`?

lib/hanami/model.rb
```diff
@@ -73,6 +73,7 @@ def self.container
 
     # @since 0.1.0
     def self.load!(&blk) # rubocop:disable Metrics/AbcSize
+      configuration.gateway.use_logger(configuration.logger)       unless configuration.logger.nil?
```
mereghost: What do you think of introducing a `NullObject`/`Mimic` logger? With ti we would not check for nils and it would simply do nothing. Yeah, one more method call, but cleaner code.

lib/hanami/model/migrator.rb
mereghost: From what I understand he will build the logger later on.

pascalbetz: Wouldn't it be easier to pass the logger? I don't see that it's the Migrators responsibility to build it. Harder to configure logger/pattern like this, no?

lib/hanami/model.rb
pascalbetz: 👍

jodosha: @mereghost @pascalbetz By avoiding the call to `use_logger`, we avoid the overhead of thousands of useless method calls against a _null object_.

Because once `gateway` receives a logger, it uses it. It wouldn't log, because it's a _null object_, but still it invokes methods on it. Which is a useless overhead.

WDYT?

lib/hanami/model/migrator.rb
jodosha: @pascalbetz @mereghost TBH, I injected that dependency only for testing purposes, but I don't like it. Also, look at how many times is passed down from method call, to method call, deep in the stack.

I can refactor it, by adding a `Configuration#migrations_logger` as private API, so we can remove it from all these method signatures **and** have an easy access to stub it during test.

lib/hanami/model.rb
pascalbetz: Wouldn't we have a logger set most of the times anyway?

TiteiKo: @pascalbetz In development, yes, but not necessarily in production

README.md
```diff
@@ -258,6 +258,25 @@ puts user.created_at.to_s # => "2016-09-19 13:40:13 UTC"
...
+Hanami::Model.configure do
+  # ...
+  logger "log/development.log", level: :debug
+end
+```
+
+It accepts the following arguments:
+
+  * `stream`: a Ruby StringIO object - it can be `$stdout` or a path to file (eg. `"log/development.log"`) - Defaults to `$stdout`
+  * `:level`: logging level - it can be: `:debug`, `:info`, `:warn`, `:info`, `:warn`, `:error`, `:fatal`, `:unknown` - Defaults to `:debug`
```
AlfonsoUceda: double `:warn` and `:info`

lib/hanami/model.rb
AlfonsoUceda: At least in production we don't log SQL queries, only the needed.

pascalbetz: I see.

Here is a benchmark, so we can make informed decisions:-)

> Comparison:
no logger, nil check:  8778993.1 i/s
   null logger, call:  6309574.3 i/s - 1.39x slower

```ruby
require 'benchmark/ips'


class NullLogger
  def info(message)
  end
end

class Something
  def initialize(logger)
    @logger = logger
  end

  def work_with_null_check
    @logger.info('working hard') if @logger
  end

  def work_without_null_check
    @logger.info('working hard')
  end
end

null_logger = Something.new(NullLogger.new)
no_logger = Something.new(nil)

Benchmark.ips do |x|
  x.compare!
  
  x.report('null logger, call') do
    null_logger.work_without_null_check
  end

  x.report('no logger, nil check') do
    no_logger.work_with_null_check
  end
  
end

```

lib/hanami/model/migrator.rb
mereghost: 👍

jodosha:
@mereghost @pascalbetz @TiteiKo @AlfonsoUceda I've refactored, according to your requests. Please review. 😄

PR #370, opened 2017-01-11
[WIP] upgrade to rom-sql 1.0.0

solnic:
### TODO

- [x] figure out why tests are failing with attributes that failed to inferred (while the same thing happens with rom-sql 0.9.1, tests are passing regardless)
- [x] figure out why entities schema coercion behaves differently (thus a couple of tests are failing)
- [x] bump rom deps to released final versions (once we have them :))
- [ ] fix remaining specs for postgresql and mysql

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,6 @@
 source 'https://rubygems.org'
 gemspec
 
-gem 'rom', git: 'https://github.com/rom-rb/rom.git', branch: 'master'
-gem 'rom-sql', git: 'https://github.com/rom-rb/rom-sql.git', branch: 'master'
-gem 'rom-repository', git: 'https://github.com/rom-rb/rom-repository.git', branch: 'master'
-
 unless ENV['TRAVIS']
   gem 'byebug', require: false, platforms: :mri
   gem 'yard',   require: false
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ unless ENV['TRAVIS']
 end
 
 gem 'hanami-utils', '~> 1.0.0.beta1', require: false, github: 'hanami/utils', branch: '1.0.x'
+gem 'sequel', git: 'https://github.com/jeremyevans/sequel.git', branch: 'master'
 
 platforms :ruby do
   gem 'sqlite3', require: false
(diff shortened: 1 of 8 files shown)


solnic:
@jodosha so, that's practically done, just waiting for final releases of rom gems. As you can see changes are tiny. I'm not 100% happy with schema mapping though. I don't have any immediate ideas how to make it work better. From what I've seen hanami entities are supposed to handle coercions both ways, I understand why you did it like that but it's not how rom works, as we have separated input/output coercions in rom 3.0.0. I don't want this to block us though, so let's just stick to what we have now and we can figure it out in hanami-model 2.0.0 (or *maybe* 1.x).

solnic:
Uhm, not sure why travis is failing, tests pass on my machine for all dbs

jodosha:
@solnic It fails for me locally too. Please remember that when you run `bundle exec rake` it only run specs agains SQLite. If you want to run for another database you should do:

```shell
DB=sqlite bundle exec rake
# or
DB=mysql bundle exec rake
# or
DB=postgresql bundle exec rake
```

solnic:
@jodosha oh damn :D ok I'll look into that :)

solnic:
@jodosha tests are failing on postgres because inferrer doesn't support `line`, `circle` and **especially** custom-made types, like the `inventory_item` one. It passes on rom-sql 0.9.x probably because we are inferring some default type by an accident for these columns.

@flash-gordon is it even possible to infer custom-defined types?

solnic:
Answering to myself: nope, because custom type has no info about any matching ruby type:

``` ruby
[:composite1, {:oid=>6663255, :db_type=>"inventory_item", :default=>"ROW('fuzzy dice'::text, 42, 1.99)", :allow_null=>true, :primary_key=>false, :type=>nil, :ruby_default=>nil}]
```

No idea what to do with that, except having ability to manually specify the attribute in schema

flash-gordon:
@solnic I didn't look deep, but I saw `Ecto` handles it, so it is possible in one way or another. Right now I can tell that Sequel has a weird API for composite types https://github.com/jeremyevans/sequel/blob/master/spec/adapters/postgres_spec.rb#L3469 I'll have a more thorough look into that.

flash-gordon:
and ofc it'll require to run more queries to system tables/views to get the info about composite types

solnic:
@flash-gordon oh that's interesting, thanks!

jodosha:
I'm closing this in favor of #377, which includes @solnic's work too. Thank you @solnic

PR #375, opened 2017-01-24
Raise exception when Repository#find is used against a table w/o primary key

### LINKED ISSUE #374: Repository find method acts weird when table has no primary_key set
I've created model without primary key with that migration:
```
Hanami::Model.migration do
  change do
    create_table :users do
      column :id, Integer
    end
  end
end
```
And then had some fun with repository
```
2.3.1 :036 > UserRepository.new.clear
 => 0 
2.3.1 :037 > UserRepository.new.create(id: 1)
 => #<User:0x007fcc2c67d658 @attributes={:id=>1}> 
2.3.1 :038 > UserRepository.new.find(1)
 => nil 
2.3.1 :039 > UserRepository.new.find(nil)
 => #<User:0x007fcc2c62ee18 @attributes={:id=>1}> 
2.3.1 :040 > UserRepository.new.create(id: 2)
 => #<User:0x007fcc2c5f7df0 @attributes={:id=>2}> 
2.3.1 :041 > UserRepository.new.find(nil)
ROM::TupleCountMismatchError: The relation consists of more than one tuple
```
It's a little weird and I hope I'll find some time to do more research on that

jodosha:
Raise `Hanami::Model::UnknownPrimaryKeyError` when `Repository#find` is used against a table w/o a primary key.

---

/cc @bbonislawski @beauby @solnic @hanami/issues for review.

Closes #374

### DIFF
--- a/lib/hanami/model/error.rb
+++ b/lib/hanami/model/error.rb
@@ -94,5 +94,14 @@ module Hanami
     # @since x.x.x
     class UnknownDatabaseTypeError < Error
     end
+
+    # Unknown primary key error
+    #
+    # @since x.x.x
+    class UnknownPrimaryKeyError < Error
+      def initialize(relation)
+        super("Can't find primary key for `#{relation}' table")
+      end
+    end
   end
 end
--- a/lib/hanami/repository.rb
+++ b/lib/hanami/repository.rb
@@ -151,7 +151,7 @@ module Hanami
     #
     # @since 0.7.0
     # @api private
-    def self.define_relation
+    def self.define_relation # rubocop:disable Metrics/MethodLength
       a = @associations
       s = @schema
 
--- a/lib/hanami/repository.rb
+++ b/lib/hanami/repository.rb
@@ -151,19 +151,13 @@ module Hanami
     #
     # @since 0.7.0
     # @api private
-    def self.define_relation # rubocop:disable Metrics/MethodLength
+    def self.define_relation
       a = @associations
 
       configuration.relation(relation) do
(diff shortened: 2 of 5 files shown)


bbonislawski:
In my opinion, raising exception wouldn't be such a bad idea. Maybe I'm wrong, but I think that method `find` should look for data by table's primary key. If table doesn't have one, result of method call is unpredictable. It can return correct data in few cases, but could fail miserably in others. 
I think that user should define own `find` method for tables without PK and decide how he wants to handle that situation

solnic:
Hanami should raise an exception when `Relation#primary_key` returns `nil` because it means that hanami is unable to establish its conventional API.

jodosha:
@solnic That means to perform `n` queries when Hanami model connects? Where `n` is the number of relations (aka database tables).

solnic:
@jodosha no, I mean that you can check it ie like that: `container.gateways[:default].relations.each { |_, r| raise 'omg' unless r.primary_key }`

solnic:
so in rom-sql 1.0.0 `by_pk` will raise a missing key error from schema when a nil was passed as PK attribute name, so at least that's an improvement as it will not result in any weird query. How exactly you can ensure that this never happens in hanami is really up to you.

jodosha:
@solnic Thank you for taking care of this and getting back.

jodosha:
Update the description according to the changes introduced by https://github.com/hanami/model/pull/375/commits/0df354d429a9827c770737372b9f3a5b104ffba6

PR #377, opened 2017-02-06
ROM 3 upgrade

jodosha:
This PR introduces compatibility with ROM 3 and an extra feature for Repository.

### Repository Schema

As of 0.7, the schema of a relation is auto inferred. That means if the database type is a Postgres `text`, the repository treat it as a Ruby `String`. There are some cases where the auto inferring fails because Hanami doesn't have a corresponding Ruby type.

For this problem we introduce `Repository.schema`.

Imagine we have a database table named `colors` which is defined like this:

```ruby
Hanami::Model.migration do
  change do
    extension :pg_enum
    create_enum :rainbow, %w(red orange yellow green blue indigo violet)

    create_table :colors do
      primary_key :id

      column :name, :rainbow, null: false

      column :created_at, DateTime, null: false
      column :updated_at, DateTime, null: false
    end
  end
end
```

Because we don't support PG enum yet, we can define the database schema for the repository, by using `Hanami::Model::Sql::Types::String` as Ruby type for it.

```ruby
class ColorRepository < Hanami::Repository
  schema do
    attribute :id,         Hanami::Model::Sql::Types::Int
    attribute :name,       Hanami::Model::Sql::Types::String
    attribute :created_at, Hanami::Model::Sql::Types::DateTime
    attribute :updated_at, Hanami::Model::Sql::Types::DateTime
  end
end
```

### DIFF
--- a/lib/hanami/model/configuration.rb
+++ b/lib/hanami/model/configuration.rb
@@ -8,7 +8,7 @@ module Hanami
     # via `Hanami::Model.configure`.
     #
     # @since 0.2.0
-    class Configuration #< ROM::Configuration
+    class Configuration
       # @since 0.7.0
       # @api private
       attr_reader :mappings
@@ -30,7 +30,6 @@ module Hanami
       def initialize(configurator)
         @backend = configurator.backend
         @url = configurator.url
-        # super(configurator.backend, configurator.url)
         @migrations        = configurator._migrations
         @schema            = configurator._schema
         @gateway_config    = configurator._gateway
@@ -124,7 +123,7 @@ module Hanami
       end
 
       def rom
-        @rom ||= ROM::Configuration.new(@backend, @url)
+        @rom ||= ROM::Configuration.new(@backend, @url, infer_relations: false)
       end
 
       def load!(repositories, &blk) # rubocop:disable Metrics/AbcSize
--- a/lib/hanami/model/configuration.rb
+++ b/lib/hanami/model/configuration.rb
@@ -8,7 +8,7 @@ module Hanami
     # via `Hanami::Model.configure`.
     #
     # @since 0.2.0
-    class Configuration < ROM::Configuration
+    class Configuration #< ROM::Configuration
       # @since 0.7.0
       # @api private
       attr_reader :mappings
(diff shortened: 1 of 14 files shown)


jodosha:
@hanami/issues @hanami/ecosystem @solnic @flash-gordon Please review 👍

lib/hanami/repository.rb
```diff
@@ -153,10 +153,15 @@ def self.container
...
     def self.define_relation # rubocop:disable Metrics/MethodLength
       a = @associations
+      s = @schema
 
       configuration.relation(relation) do
-        schema(infer: true) do
-          associations(&a) unless a.nil?
+        if s.nil?
+          schema(infer: true) do
+            associations(&a) unless a.nil?
```
mereghost: can the nil check on `a` be moved to the higher scoped `if` (`if s.nil? && a` )?

lib/hanami/model/configuration.rb
```diff
@@ -122,6 +121,32 @@ def logger=(value)
         return if value.nil?
         gateway.use_logger(@logger = value)
       end
+
+      def rom
+        @rom ||= ROM::Configuration.new(@backend, @url, infer_relations: false)
```
mereghost: From what I get `infer_relations` will default to `false` but later (`#load!`) that can get set to `true` if so desired.

jodosha: But it will be useless at that point because ROM already loaded its container.

lib/hanami/repository.rb
jodosha: @mereghost Not sure I understood exactly where and why.

lib/hanami/model/configuration.rb
mereghost: Then we are going back to explicity mappings?

lib/hanami/repository.rb
mereghost: I was thinking of not even opening the schema if there were not any associations. =) Not really that important, just curious if there was a reason to open it and then check for associations.

flash-gordon: @mereghost calling `schema(infer: true)` is required, only the block is optional

mereghost: Thanks for the clarification @flash-gordon =)

PR #382, opened 2017-02-22
Username and Password detection

mereghost:
When not using JDBC the migrator was unable to find the username and password even if they where passed on the connection string. So for running migrations on Postgres for example, you'd need to pass a `PGUSER` env variable to set the username, even if it was already provided.

The issue arised from the fact that `Connection#user` only looked up to `Connection#parsed_opts` and didn't check the raw connection string or `Sequel::Connection` too.

Note: Yeah. I duplicated the jdbc tests on a non jdbc connection. I thought that it would be the best way to guarantee the same behaviour.

### DIFF
--- a/lib/hanami/model/migrator/connection.rb
+++ b/lib/hanami/model/migrator/connection.rb
@@ -83,7 +83,7 @@ module Hanami
         # @since 0.5.0
         # @api private
         def user
-          @user ||= parsed_opt('user') || raw.user
+          @user ||= parsed_opt('user') || parsed_uri.user
         end
 
         # Returns user from DB connection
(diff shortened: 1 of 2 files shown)


mereghost:
Weird. That was a pretty random break, investigating.

mereghost:
Hmmm... the error seems related to `rom-repository` update.

/cc @jodosha @solnic I took a look at the changes in `rom-repository` and nothing seems to warrant a change in behavior for the [has_many](https://github.com/hanami/model/blob/master/lib/hanami/model/associations/has_many.rb#L144) association code. Did I miss something?

jodosha:
@mereghost I fixed it. Please check https://github.com/hanami/model/commit/384c5a1101fd362dde290b5dce8945a9a1a29a9a

/cc @solnic

PR #384, opened 2017-02-28
Parse host and port from connection url with params

artofhuman:


TiteiKo:
I don't understand the need here, @artofhuman could you explain please?

jodosha:
@artofhuman Can you please share how do you use the migrator with MySQL? Thanks 😄

artofhuman:
@TiteiKo documentation for Postgres https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING says we can use params in connection uri. I think we should support all variants.

Now if we use url like this `postgresql:///mydb?host=localhost&port=6433&user=postgres&password=testpasswd`

hanami sets only `database, user, password`, but `host` and `port` do not parse from this url and use defaults.

artofhuman:
@jodosha I'm not mysql expert ;) I read doc, and see mysql don`t support this case. May be we should implement this only for postgres adapter?

jodosha:
@artofhuman Thanks for getting back so quickly. I assumed this change was for MySQL, not for Postgres, cause I missed the `postgres://` part of the URI.

PR #391, opened 2017-03-16
WIP: Add tests for table alterations

### LINKED ISSUE #387: Strange behavior following documentation
In [documentation](http://hanamirb.org/guides/migrations/alter-table/) has a `#set_column_not_null` method, but doesn't work.

```
From: /app/db/migrations/20170304172747_add_type_to_bikes.rb @ line 6 :

     1: Hanami::Model.migration do
     2:   up do
     3:     add_column :bikes, :type, String
     4:     execute "update bikes set type = ''"
     5:     binding.pry
 =>  6:     set_column_not_null :bikes, :type
     7:   end
     8:
     9:   down do
    10:     drop_column :bikes, :type
    11:   end

[1] pry(#<Sequel::Postgres::Database>)> set_column_not_null :bikes, :type
NameError: undefined local variable or method `set_column_not_null' for #<Sequel::Postgres::Database:0x0000000378e490>
Did you mean?  set_column_default
from (pry):1:in `block (2 levels) in <top (required)>'
[2] pry(#<Sequel::Postgres::Database>)> respond_to? :set_column_not_null
=> false
[3] pry(#<Sequel::Postgres::Database>)>
```

cllns:
In the process of trying to fix #387, I noticed that none of the [table alterations](test/fixtures/database_migrations/20150612124205_table_alterations.rb) migration was being tested. And there were commented out lines from that file.

This is a work in progress.

I need to get a MySQL server running locally so I can run those tests.

I have one test failing when I run the postgres tests (`DB=postgresql bundle exec rake test:integration`)
```
  1) Failure:
Hanami::Model.migration::PostgreSQL::columns#test_0002_defines column defaults [test/integration/migration/postgresql.rb:427]:
Expected: "'-1'::integer"
  Actual: "(-1)"
```
`"(-1)"` is what's [expected on linux, but not OS X](https://github.com/cllns/hanami-model/blob/ba298001512d65b7d145d98be2df54030bcdce60/test/integration/migration/postgresql.rb#L422-L425). It seems like we can possibly get rid of the per-platform expectation, which would be 💯 

I just upgraded to macOS Sierra (10.12.3), so I'm not sure if it's related to that. Travis runs on linux, not OS. Does anyone else have insight into that?

Also, the constraints are not being created on SQLite, as far as I can tell. Not sure if that's a bug in `hanami-model` or `sequel`, or just something SQLite doesn't support.

Also need to make sure I'm testing everything that happens in this migration. Let me know what you think.

### DIFF
--- a/test/fixtures/database_migrations/20150612124205_table_alterations.rb
+++ b/test/fixtures/database_migrations/20150612124205_table_alterations.rb
@@ -1,72 +1,43 @@
 Hanami::Model.migration do
   change do
     case ENV['HANAMI_DATABASE_TYPE']
-    when 'sqlite'
-      create_table :songs do
-        column :title, String
-        column :useless, String
-
-        foreign_key :artist_id, :artists
-        index :artist_id
-
-        add_constraint(:useless_min_length) { char_length(useless) > 2 }
-      end
-
-      alter_table :songs do
-        add_primary_key :id
-
-        add_column      :downloads_count, Integer
-        set_column_type :useless,         File
-
-        rename_column :title, :primary_title
-        set_column_default :primary_title, 'Unknown title'
-
-        # add_index :album_id
-        # drop_index :artist_id
-
-        # add_foreign_key :album_id, :albums, on_delete: :cascade
-        # drop_foreign_key :artist_id
-
-        # add_constraint(:title_min_length) { char_length(title) > 2 }
-
-        # add_unique_constraint [:album_id, :title]
-
-        drop_constraint :useless_min_length
-        drop_column     :useless
-      end
     when 'postgresql'
-      create_table :songs do
-        column :title, String
-        column :useless, String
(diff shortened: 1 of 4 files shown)


cllns:
Got MySQL set up. Still have tests failing, but I think they're legitimate failures of constraints not being created, since they don't show up when I search through the generated schema manually.

What does everyone think of the current approach? Personally, I find it confusing to follow. Right now all these alterations are in the `test/fixtures/database_migrations/20150612124205_table_alterations.rb` and all migrations in `test/fixtures/database_migrations/` are run before the tests start. With them all in the same file, we can't test state of DB schema before & after, but just the outcome after :/. Instead, what do y'all think about in-lining the migrations in the test file(s), to provide more granular, specific, tests that would be easier to follow? @hanami/core

jodosha:
After a quick check with @cllns we decided to close this outdated PR.

PR #393, opened 2017-04-20
Fix Minitest nil-comparison warnings

kaikuchn:
When running the test suite the output was littered with warnings from minitest. I tracked those down and fixed the issues.

Use of `must_equal` with `nil` resulted in the following warning:
```
Use assert_nil if expecting nil from `must_equal'. This will fail in MT6.
```

### DIFF
--- a/test/entity/automatic_schema_test.rb
+++ b/test/entity/automatic_schema_test.rb
@@ -79,7 +79,7 @@ describe Hanami::Entity do
       it 'returns nil if not present in attributes' do
         entity = described_class.new
 
-        entity.id.must_equal nil
+        entity.id.must_be_nil
       end
     end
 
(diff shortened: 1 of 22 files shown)


TiteiKo:
Hello @kaikuchn 
Although we are thankful for your contribution, we currently are [migrating tests to rspec](https://github.com/hanami/model/issues/392), and so we won't merge this PR.

Don't let this bad luck discourage you, we are always welcoming contributions, and I hope I'll have the pleasure to merge your next PR ;)

kaikuchn:
Ah, sorry for not looking at the issues list! And don't worry I'm not in the least discouraged since I completely agree with the procedure in light of the transition to rspec.

Btw. should new tests for contributions then be written in rspec even though the transition has not yet begun?

PR #395, opened 2017-04-21
Symbolize entities with schema only shallow

kaikuchn:
See https://github.com/hanami/model/issues/394 for discussion.

Entities with inferred/automatic schema and entities without a schema deeply symbolise / hashify input values as before.
Manual schema however only performs shallow symbolising. It is now left to the user to ensure embedded types are correct (e.g., `Types::Hash.symbolize` for symbolized hashes).

### DIFF
--- a/lib/hanami/model/sql/entity/schema.rb
+++ b/lib/hanami/model/sql/entity/schema.rb
@@ -38,6 +38,24 @@ module Hanami
             freeze
           end
 
+          # Process attributes
+          #
+          # @param attributes [#to_hash] the attributes hash
+          #
+          # @raise [TypeError] if the process fails
+          #
+          # @since 1.0.1
+          # @api private
+          def call(attributes)
+            Utils::Hash.new(
+              schema.call(attributes)
+            ).deep_symbolize!
+          end
+
+          # @since 1.0.1
+          # @api private
+          alias [] call
+
           # Check if the attribute is known
           #
           # @param name [Symbol] the attribute name
--- a/lib/hanami/entity/schema.rb
+++ b/lib/hanami/entity/schema.rb
@@ -73,7 +73,7 @@ module Hanami
           if attributes.nil?
             {}
           else
-            attributes.dup
+            Utils::Hash.new(attributes.dup).deep_symbolize!
           end
         end
 
@@ -223,7 +223,7 @@ module Hanami
       def call(attributes)
         Utils::Hash.new(
           schema.call(attributes)
-        ).deep_symbolize!
+        ).symbolize!
       end
 
       # @since 0.7.0
--- a/test/entity/manual_schema_test.rb
+++ b/test/entity/manual_schema_test.rb
@@ -103,6 +103,31 @@ describe Hanami::Entity do
 
         exception.message.must_equal '"foo" (String) has invalid type for :code'
       end
+
+      it 'symbolizes nested hash keys according to schema' do
+        entity = PageVisit.new(
+          id: 42,
+          start: DateTime.now,
+          end: (Time.now + 53).to_datetime,
+          visitor: {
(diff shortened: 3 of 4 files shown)


lib/hanami/model/sql/entity/schema.rb
```diff
@@ -38,6 +38,24 @@ def initialize(registry, relation, mapping)
...
+          # @param attributes [#to_hash] the attributes hash
+          #
+          # @raise [TypeError] if the process fails
+          #
+          # @since 1.0.1
+          # @api private
+          def call(attributes)
+            Utils::Hash.new(
+              schema.call(attributes)
+            ).deep_symbolize!
```
mereghost: Quick question: why is this needed? Which cases does it cover?

kaikuchn:
This is for automatic schema. When the Schema is inferred from the
relation, a sql entity schema is built. In that case we want to deeply
symbolise, right?
Marcello Rocha <notifications@github.com> schrieb am Fr. 21. Apr. 2017 um
17:59:

lib/hanami/model/sql/entity/schema.rb
kaikuchn: Before this change `Sql::Entity::Schema#call` was inherited from `Entity::Schema` which had this exact implementation. Since I changed the latter but did not want to change the behaviour of the former, I added this to override the inherited `Entity::Schema#call` method. I.e., for inferred schema the behaviour should be as it was before.

I did not ponder how inferred schemas treat associations at all. 🤔  Maybe we should first discuss the desired behaviour for all the cases?

jodosha:
@kaikuchn Thanks for this PR. Before to proceed, let me to provide a short background.

---

`Utils::Hash#deep_symbolize!` is used to make sure that deep nested hashes' keys are converted from strings to symbols.

Now the problem is that `Utils::Hash` is too aggressive in its operations: it tries to convert all the objects that respond to `#to_hash` into `Hash`/`Utils::Hash` objects.

This is the source of our problem: because `Entity` responds to `#to_hash`, it gets serialised.

---

Now, if we stop using `deep_symbolize!` in favor of `symbolize!` we won't be able to symbolize deep nested hashes anymore, which will probably introduce other problems down the path.

---

I wrote a [proposal](https://discourse.hanamirb.org/t/transform-utils-string-and-utils-hash-to-modules/336) on how to move in the right direction in the future. Please let me know what you think.

As part of that proposal, I implemented `Utils::Hash.deep_symbolize` with has a less aggressive coercion policy: it will deep symbolize hashes and leave other objects like `Hanami::Entity` as they are. Please review https://github.com/hanami/utils/pull/210

kaikuchn:
I like that proposal a lot! It addresses the root of the problem instead of working around it. IIRC this isn't the first time that the aggressive hashifying of `deep_symbolize` has caused trouble.

jodosha:
@kaikuchn 
> IIRC this isn't the first time that the aggressive hashifying of deep_symbolize has caused trouble.

you're right.

lib/hanami/model/sql/entity/schema.rb
mereghost: Sorry for the delay, RL got in the way. =/
 It's as @jodosha said, `deep_symbolize!` ends up being way too aggressive and it ends up calling `to_hash` on an entity when it's returned by the association.

It'd remove this method as it would probably ending creating more trouble while we do not "fix" `utils`.

marciol: Hi all, I'm suffering with this call to deep_symbolize! right now. Nothing that I do solve this issue and I'm forcing myself to `stringify!` on consumers of my entitties.

kaikuchn: Hi Marciol, what exactly is your issue? Maybe join us in the Chat, somebody will be able to help out :)

Issue #392, opened 2017-03-20
Migrate from Minitest to RSpec


jodosha:
A few rules to follow:

  * If you want to work on this, please leave a comment here. We want to avoid many people to work on the same task.

  * We expect this migration to happen within a week from now.

  * Do not change the tests, only migrate from Minitest to RSpec syntax. If not sure, please ask here.

  * Do not remove the old tests, only add the new ones.

  * Make CI build to use RSpec tests.

  * **Target `develop` branch, instead of `master`.**

  * All the tests artefacts like fixtures, goes under `spec/support`.

  * Try to reflect the structure under `lib/`. Eg. `lib/hanami/view.rb` should be tested under `spec/hanami/view_spec.rb`, while `test/rendering_test.rb` should go under `spec/integration/rendering_spec.rb`.

If you want to take inspiration, please have a look at: https://github.com/hanami/mailer/pull/67

tejasbubane:
I would like to work on this.

AlfonsoUceda:
@tejasbubane go ahead ;)

zenspider:
Can you please record some baseline numbers from minitest (probably from the CI) such that you can compare to the rspec version? # tests run, % passed, # assertions/expections, and time to run (and memory RSS if you can!)?

jodosha:
@zenspider Yes, of course.

morrme:
@tejasbubane  Are you still working on this task?  How's it going?

tejasbubane:
Working on it. Was on holidays for a few days. Will send a PR soon.

jodosha:
@tejasbubane ping.

jodosha:
Implemented by #396

PR #396, opened 2017-04-21
Migrate to rspec 

bbonislawski:
Work for #392 

I'm not sure about my way of including examples for database-specific migrations. RSpec doesn't allow you to load specs with `load` AFAIK. 

Please correct me if I'm wrong and there is better way to do that

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,2 @@
+--color
+--require spec_helper
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,7 @@ Lint/IneffectiveAccessModifier:
 Metrics/BlockLength:
   Exclude:
     - "test/**/*.rb" # Minitest syntax generates this false positive
+    - "spec/**/*.rb"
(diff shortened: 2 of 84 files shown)


jodosha:
@bbonislawski Hi and thanks for this PR. Which specs are still missing?

bbonislawski:
@jodosha I didn't mean there are missing specs. I'm not sure if https://github.com/hanami/model/pull/396/files#diff-9822821bd605b402d1e2e6e2a28989db 
this is good solution for the problem and I'd like to get some feedback about it

jodosha:
@bbonislawski I've finally merged this. Thank you again! 👍

solnic:
YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES YES

bbonislawski:
It was pleasure!

PR #397, opened 2017-04-23
Changing example in code documentation

rogeriozambon:
Hey,

I was reading the Entity and Repository code documentation and realized that the instance of an Entity class was wrong. In the documentation, the example guides as follows:

```ruby
2.3.4 (main)> class User
2.3.4 (main)>   include Hanami::Entity
2.3.4 (main)> end  
TypeError: wrong argument type Class (expected Module)
from (pry):2:in `include'
```

So I changed the documentation. Just it :smile:
Thanks guys.

### DIFF
--- a/lib/hanami/entity.rb
+++ b/lib/hanami/entity.rb
@@ -19,8 +19,7 @@ module Hanami
   # @example With Hanami::Entity
   #   require 'hanami/model'
   #
-  #   class Person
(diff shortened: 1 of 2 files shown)


cllns:
great find. thank you! The reason this was there was because `Hanami::Entity` used to be a module :)

PR #400, opened 2017-05-04
Support for rom-sql 1.3 and dry-types 0.10

jodosha:
The recent changes introduced by `rom-sql` 1.3 and `dry-types` 0.10 led to a failure in our build. When an inferred **and** mapped schema are used together, the attributes of an entity weren't coerced.

Example:

```ruby
class Operator < Hanami::Entity
end

class OperatorRepository < Hanami::Repository
  self.relation = :t_operator

  mapping do
    attribute :id,   from: :operator_id
    attribute :name, from: :s_name
  end
end

operator = Operator.new(name: :foo)
operator.name
# => :foo
# WRONG, expected "foo"
```

---

After a long chat with @flash-gordon (thank you again, man), I decided to bump the dependencies of `rom-sql` and `dry-types` for us and to us a new API from `dry-types` to internally convert database types into DRY/Hanami types.

---

/cc @hanami/core

### DIFF
--- a/lib/hanami/model/sql/types.rb
+++ b/lib/hanami/model/sql/types.rb
@@ -34,26 +34,26 @@ module Hanami
           # @since 0.7.0
           # @api private
           MAPPING = {
-            Types::String.with(meta: {})            => Schema::String,
-            Types::Int.with(meta: {})               => Schema::Int,
-            Types::Float.with(meta: {})             => Schema::Float,
-            Types::Decimal.with(meta: {})           => Schema::Decimal,
-            Types::Bool.with(meta: {})              => Schema::Bool,
-            Types::Date.with(meta: {})              => Schema::Date,
-            Types::DateTime.with(meta: {})          => Schema::DateTime,
-            Types::Time.with(meta: {})              => Schema::Time,
-            Types::Array.with(meta: {})             => Schema::Array,
-            Types::Hash.with(meta: {})              => Schema::Hash,
-            Types::String.optional.with(meta: {})   => Schema::String,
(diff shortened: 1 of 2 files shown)


lib/hanami/model/sql/types.rb
```diff
@@ -34,26 +34,26 @@ module Schema
...
+            Types::Int.pristine               => Schema::Int,
+            Types::Float.pristine             => Schema::Float,
+            Types::Decimal.pristine           => Schema::Decimal,
+            Types::Bool.pristine              => Schema::Bool,
+            Types::Date.pristine              => Schema::Date,
+            Types::DateTime.pristine          => Schema::DateTime,
+            Types::Time.pristine              => Schema::Time,
+            Types::Array.pristine             => Schema::Array,
+            Types::Hash.pristine              => Schema::Hash,
+            Types::String.optional.pristine   => Schema::String,
```
flash-gordon: @jodosha I'm not sure if we need this now

jodosha: @flash-gordon What is "this", specifically? 😉  `#pristine`? `#optional` types? The whole mapping?

flash-gordon: sorry, all the optionals :) since you unwrap them before calling `.fetch`

jodosha: @flash-gordon I see. By returning `unwrapped` (aka skipping `MAPPING.fetch`) I get three failures.

By debugging this scenario, I noticed that `unwrapped` still doesn't coerce values.

jodosha:
Merging this as it fixes the build for all the branches and PRs.

PR #405, opened 2017-05-19
Do not override class-level input in commands (fixes #404)

### LINKED ISSUE #404: Stack Overflow on thousands `update` call
You could try this code block.

```ruby
repo = FooRepository.new
3700.times { repo.update(1, bar: 1) }
```

This code got Stack Overflow when it reached around 3650th iteration.
Assuming that there are 3 Hanami customized methods in the update method, which is subclass of ROM::Repo. and That methods was getting propagated every time update method had been called.

flash-gordon:
The previous version modified class-level state on each command call. This led to a big performance impact and ended up with a StackOverfow exception. The new code still builds inputs on every request but this can be optimized in future.

### DIFF
--- a/lib/hanami/model/plugins/mapping.rb
+++ b/lib/hanami/model/plugins/mapping.rb
@@ -37,8 +37,8 @@ module Hanami
           # @since 0.7.0
           # @api private
           def build(relation, options = {})
-            input(InputWithMapping.new(relation, input))
-            super(relation, options.merge(input: input))
+            wrapped_input = InputWithMapping.new(relation, options.fetch(:input) { input })
+            super(relation, options.merge(input: wrapped_input))
           end
         end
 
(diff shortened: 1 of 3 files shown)


solnic:
Where was this happening? For some reason I can't see it in the diff

flash-gordon:
@solnic with every `UpdateCommand.build` call. Calling `Command.input(new_input(input))` sets class-level `input` and with every call it gets bigger because previous input is stored inside a new one.

solnic:
@flash-gordon ahh 💩 right. that's why I've been thinking about freezing classes on finalization...something to do in rom 4.0 :)

flash-gordon:
yes, good idea, however it's not that easy, command classes can be created by the command compiler after finalization

solnic:
@flash-gordon auto-commands should be frozen too, I'll keep that in mind when working on rom 4.0

jodosha:
@flash-gordon Thank you very much for this PR. 💯

Issue #412, opened 2017-07-12
Automated setting of created_at does not work


BenBach:
I have a repository with a created_at column with `Null: false` constraint. My database is postgresql

When I try to do a create similar to the example code from the guide it fails because created_at can't be null (http://hanamirb.org/guides/models/repositories/)

```
repository = BookRepository.new
book = repository.create(title: "Hanami")
```

Expected: `created_at` is set automatically

BenBach:
I found the problem:

If only a created_at column is existing this does not work. Once I add the updated_at column it works
I think having only a created at column should work as well (eg. in case of immutable objects)

TiteiKo:
Hi @BenBach, thanks for reporting this!

I've made a quick fix, could you review it please? :)

BenBach:
@TiteiKo I did. Thanks for fixing it so quick. :-)

jodosha:
Fixed by #413

PR #413, opened 2017-07-12
Fix timestamps for models not having updated_at column.

TiteiKo:
Fix for #412 

(full disclosure: this is a quick fix, `value[:updated_at]` will still be set, but, as it's not in the schema, won't be present in the returned object)

### DIFF
--- a/lib/hanami/model/plugins/timestamps.rb
+++ b/lib/hanami/model/plugins/timestamps.rb
@@ -24,7 +24,7 @@ module Hanami
           def initialize(relation, input)
             super
             columns     = relation.columns.sort
-            @timestamps = (columns & TIMESTAMPS).count.positive?
+            @timestamps = columns & TIMESTAMPS
           end
 
           # Processes the input
@@ -49,7 +49,7 @@ module Hanami
           # @since 0.7.0
           # @api private
           def timestamps?
-            @timestamps
+            !@timestamps.empty?
           end
         end
 
@@ -63,7 +63,7 @@ module Hanami
           # @since 0.7.0
           # @api private
           def _touch(value, now)
-            value[:updated_at] ||= now
+            value[:updated_at] ||= now if @timestamps.include?(:updated_at)
             value
           end
         end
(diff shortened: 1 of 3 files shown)


lib/hanami/model/plugins/timestamps.rb
```diff
@@ -24,7 +24,7 @@ class InputWithTimestamp < WrappingInput
           def initialize(relation, input)
             super
             columns     = relation.columns.sort
-            @timestamps = (columns & TIMESTAMPS) == TIMESTAMPS
+            @timestamps = (columns & TIMESTAMPS).count.positive?
```
pettomartino: isn't `!(columns & TIMESTAMPS).empty?` more efficient?

And just for the sake of knowledge let me ask you,  I know it was previously there but isn't better to have the boolean check (either `.count.positive?` or `.empty?` placed in `timestamp?` method instead of in the initializer?

TiteiKo:
@pettomartino Yep, you're wright. I think part of my brain was still asleep yesterday.

So, this PR is now a nice PR and not a quick (and dirty) bugfix: no setting up updated_at if it doesn't exist :sparkles: 

@pettomartino @BenBach Please review again?
ping @hanami/core

lib/hanami/model/plugins/timestamps.rb
```diff
@@ -24,7 +24,7 @@ class InputWithTimestamp < WrappingInput
           def initialize(relation, input)
             super
             columns     = relation.columns.sort
```
TiteiKo: What's the point of sorting the columns here, or even instantiating `columns`? :confused:

pettomartino: Previously when it was `== TIMESTAMPS` it was important. There's no need anymore. 
You could also remove the temp var
`@timestamps = relation.columns & TIMESTAMPS`

TiteiKo: Yep, that's what I meant by "what's the point of [...] instantiating `columns`" ;)

Anyway, that's cleaned up too

PR #415, opened 2017-07-17
handles special characters

olistik:
The ENV variable `PGPASSWORD` needs a string that is not percent-encoded, even though the official documentation states otherwise:

https://www.postgresql.org/docs/current/static/libpq-envars.html

> PGPASSWORD behaves the same as the password connection parameter.

https://www.postgresql.org/docs/current/static/libpq-connect.html

> Percent-encoding may be used to include symbols with special meaning in any of the URI parts.

Fixes https://github.com/hanami/hanami/issues/804

### DIFF
--- a/lib/hanami/model/sql/consoles/postgresql.rb
+++ b/lib/hanami/model/sql/consoles/postgresql.rb
@@ -59,7 +59,7 @@ module Hanami
           # @since 0.7.0
           # @api private
           def configure_password
-            ENV[PASSWORD] = @uri.password unless @uri.password.nil?
+            ENV[PASSWORD] = URI.decode(@uri.password) unless @uri.password.nil?
           end
         end
       end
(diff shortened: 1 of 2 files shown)


TiteiKo:
Do you think you could add a test to prevent regressions?

olistik:
I’m having a problem running the tests..
I added this:

```ruby
    context 'when the password contains percent encoded characters' do
      let(:uri) { URI.parse('postgres://username:p%40ss@localhost:1234/foo_development') }

      it 'sets the PGPASSWORD environment variable decoding special characters' do
        console.connection_string
        expect(ENV['PGPASSWORD']).to eq('WRONG')
        ENV.delete('PGPASSWORD')
      end
    end
```

to `spec/unit/hanami/model/sql/console/postgresql.rb`

and then ran:

- `rspec`
- `bundle exec rake spec:all`
- `rspec spec/unit/hanami/model/sql/console/postgresql.rb`

but in each case I can’t see the expected failure.

Am I missing something obvious?

TiteiKo:
`DB=postgresql bundle exec rake spec:all`

olistik:
@TiteiKo let me know if the provided test is enough or if you'd like to cover more cases :-)

TiteiKo (approved):

AlfonsoUceda (approved):
perfect @olistik, great investigation and thanks for the PR

mereghost (approved):

PR #419, opened 2017-07-25
Feature: Support for many-to-many associations

mereghost:
# Support for Many-To-Many

Many-to-many associations offer some distinct challenges, the main one is that there's an association table involved.

```sql lite
CREATE TABLE books (
  id integer PRIMARY KEY
  title text NOT NULL
)

CREATE TABLE categories (
  id integer PRIMARY KEY
  name text NOT NULL
)

CREATE TABLE books_categories (
  id integer PRIMARY KEY
  book_id integer REFERENCES books(id) ON DELETE CASCADE
  category_id integer REFERENCES categories(id) ON DELETE CASCADE
)
```

Also a frequently you'd want to store aditional data on the association record, like timestamps and whatnot.

```ruby
class Book < Hanami::Entity ; end
class Category < Hanami::Entity ; end
class BooksCategories < Hanami::Entity ; end # association entity

class BookRepository < Hanami::Repository
  associations do
    has_many :categories, through: :books_categories
  end
end

class CategoryRepository < Hanami::Repository
  associations do
    has_many :books, through: :books_categories
  end
end

class BooksCategoriesRepository < Hanami::Repository # association repository
  associations do
    belongs_to :books
    belongs_to :categories
  end
end
```

## What operations will we offer

- Retrieving the associated records (`#assoc`)
- Retrieving all associated records as an Array of Entities (`#to_a`)
- Iterating through the associated records (`#each`)
- Map operations through the associated records (`#map`)

These 4 operations obey the rules set by our, already shipped, `has_many` association implementation.

- Creating the association (`#add`)
- Deleting the association (`#remove`)

These 2 differ somewhat and there's no point in `update` this kind of association.

### `#add(*data)`

Contrary to the `has_many` `#add`, this won't create a record on the associated table (as it would make little sense at first) but instead create records on the association table returning those records.

`data ` here is either one or a list of entites/hashes with `:id` keys.

### `#remove(entity)`

While the `has_many` implementation merely nulls the association foreign key, here it might not be an options as it may lead to weird results (4 associations  but only 3 objects), so we delete the association record.

For this we first query the table for the `id` of the association record (using the `through` relation) and then we delete it.

### Why `has_many :through`?	

The main reason is that ROM does not support out of the box `has_and_belongs_to_many`, so we could implement this, but would be all our code to support.

Also IMO `habtm` is rarely used in the wild as we usually need to record some metadata with the association table. With that in mind, `#add` and `#remove` would be made useless, but they cover the `habtm` usage somewhat.


mereghost:
Also, I tried to hide the association relation and entity. I failed. 😭

.gitignore
```diff
@@ -19,3 +19,4 @@ spec/reports
 test/tmp
 test/version_tmp
 tmp
+.idea
```
cllns: mysterious 😮

cllns:
Great work!

I think `has_and_belongs_to_many` is an anti-pattern, and only supporting `has_many :through` is more in line with Hanami's philosophy :)

Just want to clarify, `BooksCategoriesRepository` doesn't need to have that exact name, right? It could be `BookCategoryAssignmentsRepository`, for example. It computes the class name based on the `:through` key, not the names of the repositories. Is that right?

.gitignore
mereghost: I'm hiding my ideas. ;)

On a more serious tone, I frequently use Jetbrains RubyMine while doing some pretty hardcore stuff like big refactors, debugging using a, well, debugger (usually I'm a `p` debugger) and it creates this frigging directory in the root of the project. =)

mereghost:
Yeah. The name of the repository is irrelevant, I just used the current convention to not throw in more info. 

```sql
CREATE TABLE book_ontologies (
  id integer PRIMARY KEY
  book_id integer REFERENCES books(id) ON DELETE CASCADE
  category_id integer REFERENCES categories(id) ON DELETE CASCADE
)
```

```ruby
class BookOntology < Hanami::Entity ; end

class BookRepository < Hanami::Repository
  associations do ;  has_many :categories, through: :book_ontologies ; end
end

class BookOntology < Hanami::Repository
  associations do ; belongs_to :books ; belongs_to :categories ; end
end
```
This should work. The parameter for `through` is the relation name for the association relation.

.gitignore
cllns:
```
git config --global core.excludesfile ~/.gitignore_global # if you don't already have one
echo ".idea" >> ~/.gitignore_global
```
:)

mereghost: =D I still need to get my new laptop up to speed with my dotfiles.

But TBH I just commited this by mistake. xD I'll revert later.

mereghost:
@hanami/core anyone against merging this into **develop**?

jodosha (approved):
@mereghost The code LGTM. I haven't tried it yet in a project, but will do soon. 👍 TYVM for this feature! 💯

jodosha:
@mereghost I'd like to merge the latest fix in master, rebase develop and then merge this.

PR #421, opened 2017-08-01
Don't remove migrations if `Migrator#apply` fails to dump the database

### LINKED ISSUE #408: DB apply command removes migrations even if fail
Hi,

It isn't major bug but I had following inconsistency on my local environment:

```
pg_dump: server version: 9.6.3; pg_dump version: 9.5.5
pg_dump: aborting because of server version mismatch
pg_dump: server version: 9.6.3; pg_dump version: 9.5.5
pg_dump: aborting because of server version mismatch
```

When I used the `bundle exec hanami db apply` command it didn't filled up the `schema.sql` file but still removed all my migrations. Fortunately we have git ;P

jodosha:
Fixes #408

### DIFF
--- a/lib/hanami/model/migrator/adapter.rb
+++ b/lib/hanami/model/migrator/adapter.rb
@@ -133,12 +133,6 @@ module Hanami
         def new_connection(global: false)
           uri = global ? connection.global_uri : connection.uri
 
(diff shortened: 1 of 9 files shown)


jodosha:
Closed in favor of #424

PR #422, opened 2017-08-02
Allow array as Postgres JSONB values

### LINKED ISSUE #418: ArgumentError: invalid value for Hash() when working with jsonb arrays
Hi!
Looks like using automatic schema and jsonb arrays in postgres is currently not possible.

Assuming this simple setup:
```ruby
# ...create_articles.rb
Hanami::Model.migration do
  change do
    create_table :articles do
      # ...
      column :tags, 'jsonb', null: false, default: [].to_json
    end
  end
end

# ../entities/article.rb
class Article < Hanami::Entity
end

# ../repositories/article_repository.rb
class ArticleRepository < Hanami::Repository
end
```

I get a ArgumentError when trying to create a record with a jsonb array:
```ruby
repository = ArticleRepository.new
repository.create(tags: ['foo', 'bar'])
# => ArgumentError: invalid value for Hash(): ["foo", "bar"]
```

After some digging and a lot of `bundle open`_ing_ I'm pretty sure this line is to blame: https://github.com/hanami/model/blob/36003ff50c571574269cd81d0c0554cc9d949bf5/lib/hanami/model/sql/types.rb#L76, basically all jsonb columns are treated as hashes, which is what causes the exception.

The workarounds that I was able to find are:

1. Not using automatic schema (sad face)
2. This horrible hack to manually set the column type, that uses at least 2 levels of private APIs:

```ruby
class Article < Hanami::Entity
  def self.schema=(schema)
    schema.send(:schema).type.options[:member_types][:tags] = Types::Json::Array
    super
  end
end
```

_sidenote:_ I wonder if https://github.com/hanami/model/issues/385 was a final decision? Having a public API for the above seems pretty useful.

I'd be up for helping to fix this (assuming it's a bug), but I've got very limited experience with hanami and rom, so might need some guidance to do so...

jodosha:
Fix #418

### DIFF
--- a/lib/hanami/model/sql/types.rb
+++ b/lib/hanami/model/sql/types.rb
@@ -31,6 +31,19 @@ module Hanami
           Array    = Types::Strict::Nil | Types::Array.constructor(Coercions.method(:array))
           Hash     = Types::Strict::Nil | Types::Hash.constructor(Coercions.method(:hash))
 
+          JSONB = lambda do |input|
+            case input
+            when ->(i) { i.respond_to?(:to_hash) }
+              Schema::Hash.call(input)
+            when ->(i) { i.respond_to?(:to_a) }
+              Schema::Array.call(input)
+            when ::NilClass
+              input
+            else
+              raise ArgumentError.new("Exception: invalid value for JSONB(): #{input.inspect}")
+            end
+          end
+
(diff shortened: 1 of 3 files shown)


jodosha:
Closing in favor of #423

PR #423, opened 2017-08-02
Add proper handling for json and jsonb in Postgres

katafrakt:
Following #418 I discovered following problems:

- For JSONB columns it is indeed assumed they are `Hash`, not `Array`
- JSON columns do not work well - they return `Sequel::Postgres::JSONHash` instead of simple `Hash`

This PR fixes those.

### DIFF
--- a/lib/hanami/model/sql/types/schema/coercions.rb
+++ b/lib/hanami/model/sql/types/schema/coercions.rb
@@ -192,6 +192,12 @@ module Hanami
                 raise ArgumentError.new("invalid value for Hash(): #{arg.inspect}")
               end
             end
+
+            def self.json(arg)
+              hash(arg)
+            rescue ArgumentError
+              array(arg)
+            end
           end
           # rubocop:enable Metrics/MethodLength
         end
--- a/lib/hanami/model/sql/types/schema/coercions.rb
+++ b/lib/hanami/model/sql/types/schema/coercions.rb
@@ -193,10 +193,17 @@ module Hanami
               end
             end
 
+            # Coerces given argument to appropriate JSON type, i.e. Hash or Array
+            #
+            # @param arg the argument to coerce
+            #
+            # @return [Hash, Array] the result of the coercion
+            #
+            # @raise [ArgumentError] if the coercion fails
+            #
+            # @api private
             def self.json(arg)
-              hash(arg)
-            rescue ArgumentError
-              array(arg)
+              arg.respond_to?(:to_hash) ? hash(arg) : array(arg)
             end
           end
           # rubocop:enable Metrics/MethodLength
--- a/lib/hanami/model/sql/types.rb
+++ b/lib/hanami/model/sql/types.rb
@@ -74,14 +74,17 @@ module Hanami
             #
             #  MAPPING.fetch(unwrapped.pristine, attribute)
             MAPPING.fetch(unwrapped.pristine) do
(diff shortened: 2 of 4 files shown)


katafrakt:
Huh, I haven't noticed #422 being already opened, so maybe this is no longer relevant. On one hand, it fixes more, but on the other - the other one is by @jodosha so it's probably better 😉

lib/hanami/model/sql/types/schema/coercions.rb
```diff
@@ -192,6 +192,12 @@ def self.hash(arg)
                 raise ArgumentError.new("invalid value for Hash(): #{arg.inspect}")
               end
             end
+
+            def self.json(arg)
```
jodosha: @katafrakt What is this about? When is it called?

jodosha:
@katafrakt Thanks for this PR. I think yours is better than mine. As long we can quickly iterate on this, I prefer to merge it and close (unmerged) mine.

---

This PR is failing because of Rubocop checks. Please run `bundle exec rubocop -a` to automatically fix those.

---

Also, `master` branch CI build is failing because of a Travis CI upgrade of SQLite and MySQL. I'm trying to fix it as I'm typing. I'll probably ask to rebase `master` later today.

lib/hanami/model/sql/types/schema/coercions.rb
```diff
@@ -192,6 +192,12 @@ def self.hash(arg)
...
+              hash(arg)
+            rescue ArgumentError
```
jodosha: This is conceptually wrong and a magnitude slower than checking the current object. Can you please use a code similar to this? https://github.com/hanami/model/pull/422/files#diff-92790045433e0619f41285e4ad4b4af0R34 Thanks!

lib/hanami/model/sql/types/schema/coercions.rb
katafrakt: It's called [here](https://github.com/hanami/model/pull/423/files#diff-92790045433e0619f41285e4ad4b4af0R34):

```
JSON     = Types::Strict::Nil | Types::Hash.constructor(Coercions.method(:json)) | Types::Array.constructor(Coercions.method(:json))
```

Without that constructors are raising errors because of wrong data type.

lib/hanami/model/sql/types/schema/coercions.rb
katafrakt: Just to understand you correctly: is using `rescue` conceptually wrong or having this `json` coercion?

jodosha: In general, it's conceptually wrong to use exceptions for control flow.

lib/hanami/model/sql/types/schema/coercions.rb
jodosha: @katafrakt Oh right, sorry for the silly question.

I'm wondering if this is correct. `.json` can return both a `Hash` or an `Array`, according to the input. Is there the risk that `Types::Hash.constructor(Coercions.method(:json))` will receive an array?

katafrakt: @jodosha It's almost certain that `Types::Hash.constructor(Coercions.method(:json))` will receive an array at some point and that's why I decided to write this `json` coercion in the first place.

The coercion tries to convert input to `Hash` and only when it fails it tries to convert to `Array`. And I lazily used exception for that to avoid copying coercion logic, but you are right that it's not a good way to go.

katafrakt:
@jodosha PR is updated, Rubocop pleased and Travis build passed.

PR #425, opened 2017-08-05
Feature: Support for one-to-one associations

mereghost:
# Support for One-To-One

Without further ado I present you: One-to-One associations support

```sql lite
CREATE TABLE avatars (
  id integer PRIMARY KEY
  url text NOT NULL
  user_id integer REFERENCES users(id) ON DELETE CASCADE
)

CREATE TABLE users (
  id integer PRIMARY KEY
  name text NOT NULL
)
```

Declaring one is pretty simple:

```ruby
class Avatar < Hanami::Entity ; end
class User < Hanami::Entity ; end

class AvatarRepository < Hanami::Repository
  associations do
    belongs_to :users
  end
end

  associations do
    has_one :avatar
  end
end
```

## What operations will we offer

- Retrieving the associated records (`#assoc`)
- Retrieving the associated record as an entity (`#one`)

These 4 operations obey the rules set by our, already shipped, `has_many` association implementation.
- Adding a child record to an already existing parent (`#add`)
- Creating the parent-child graph on a single call (`#create`)
- Updating the associated record (`#update`)
- Removing the associated record (`#remove`)
- Replacing the associated record (`#replace`), this will perform a delete of the current associated record and create a new one.

### The conundrum of `#add` 

By default `rom` does not remove the old records when you substitute the associated record, resulting in `UNIQUE CONSTRAINT` violations, if there's such constraint, or multiple associated records if there isn't.

What did I do to fix this? Nothing yet (and that's why it's a **WIP**), but further discussion is happening on [this rom-sql issue](https://github.com/rom-rb/rom-sql/pull/226)

Please @hanami/core and community, chime in. =)

### DIFF
--- a/lib/hanami/model/association.rb
+++ b/lib/hanami/model/association.rb
@@ -1,6 +1,7 @@
 require 'rom-sql'
-require 'hanami/model/associations/has_many'
 require 'hanami/model/associations/belongs_to'
+require 'hanami/model/associations/has_many'
+require 'hanami/model/associations/has_one'
 require 'hanami/model/associations/many_to_many'
 
 module Hanami
@@ -23,10 +24,13 @@ module Hanami
       #
       # @since 0.7.0
       # @api private
+      # rubocop:disable Metrics/MethodLength
       def self.lookup(association)
         case association
         when ROM::SQL::Association::ManyToMany
           Associations::ManyToMany
+        when ROM::SQL::Association::OneToOne
+          Associations::HasOne
         when ROM::SQL::Association::OneToMany
           Associations::HasMany
         when ROM::SQL::Association::ManyToOne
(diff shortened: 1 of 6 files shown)


jodosha (commented):
@mereghost Thanks for this PR. Can you please add the specs for:

  1. Replacing the avatar (skipped spec because of https://github.com/rom-rb/rom-sql/pull/226)
  2. Removing the avatar

TYVM

mereghost:
Sure. I'm still polishing this so rest assured that they will be there. =)

mereghost:
@jodosha Not sure what's best to do: issue an update to the target relation setting the FK to `nil` or just plainly removing the record.

I'm going with the second approach for this kind of association as it makes more sense since an 1-to-1 implies that the child record has no life of its own.

jodosha (approved):

jodosha:
@mereghost Thank you for the perfect timing. 💯

PR #431, opened 2017-08-08
Allow to specify the behavior of the manual schema for an entity

### LINKED ISSUE #411: How to make Hanami::Entity behave like Dry::Struct?
For example, I have classes A and B:
```
class A < Dry::Struct
   attribute :id, Types::Strict:Int
end

class B < Hanami::Entity
  attributes do
     attribute :id, Types::Strict:Int
  end
end
```
If I create an empty instance of class A there will be an error:
```
> A.new
Dry::Struct::Error: [A.new] :id is missing in Hash input
```
But if I create an empty instance of class B - no:
```
> B.new
=> #<B:0x007f95105b8658 @attributes={}>
```

How can I get the same error for class B?
Thanks

jodosha:
This is a proposal to allow developers to specify the behavior of manual schema for entities.

## Existent behavior

```ruby
class Person < Hanami::Entity
  attributes do
    attribute :id, Types::Int
    attribute :name, Types::String
  end
end

Person.new
=> #<Person:0x007ff859e34118 @attributes={}>

Person.new(id: 1)
# => #<Person:0x007ff85acbcfc8 @attributes={:id=>1}>

Person.new(id: "1")
# => #<Person:0x007ff85a04d558 @attributes={:id=>"1"}>

Person.new(id: 1, name: "Luca")
# => #<Person:0x007ff85ab20200 @attributes={:id=>1, :name=>"Luca"}>

Person.new(id: 1, name: "Luca", foo: "bar")
# => #<Person:0x007ff859e44ea0 @attributes={:id=>1, :name=>"Luca"}>

Person.new(foo: "bar")
# => #<Person:0x007ff859e4d1e0 @attributes={}>
```

## Additional behavior

```ruby
class Person < Hanami::Entity
  attributes :strict do
    attribute :id, Types::Strict::Int
    attribute :name, Types::Strict::String
  end
end

Person.new
# => ArgumentError: :id is missing in Hash input

Person.new(id: 1)
# => ArgumentError: :name is missing in Hash input

Person.new(id: 1, name: "Luca")
# => #<Person:0x007f8476816c88 @attributes={:id=>1, :name=>"Luca"}>

Person.new(id: 1, name: "Luca", foo: "bar")
# => ArgumentError: unexpected keys [:foo] in Hash input

Person.new(foo: "bar")
# => ArgumentError: unexpected keys [:foo] in Hash input

Person.new(id: "1", name: "Luca")
# => TypeError: "1" (String) has invalid type for :id violates constraints (type?(Integer, "1") failed)
```

⚠️ **Please note that we are NOT replacing the existing behavior, we're adding additional ones, the one above is just an example.** ⚠️ 

## Allowed types

  * omitted (defaults to `:schema`, which is the **existing behavior**)
  * `:schema`
  * `:strict`
  * `:weak`
  * `:permissive`
  * `:strict_with_defaults`
  * `:symbolized`

Read more at: http://dry-rb.org/gems/dry-types/hash-schemas/

---

Closes #411

### DIFF
--- a/lib/hanami/entity/schema.rb
+++ b/lib/hanami/entity/schema.rb
@@ -175,7 +175,7 @@ module Hanami
         # @param attributes [#to_hash] the attributes hash
         #
         # @raise [TypeError] if the process fails
-        # @raise [ArgumentError] if data is missing
+        # @raise [ArgumentError] if data is missing, or unknown keys are given
         #
         # @since 0.7.0
         # @api private
@@ -183,7 +183,7 @@ module Hanami
           schema.call(attributes)
         rescue Dry::Types::SchemaError => e
           raise TypeError.new(e.message)
-        rescue Dry::Types::MissingKeyError => e
+        rescue Dry::Types::MissingKeyError, Dry::Types::UnknownKeysError => e
           raise ArgumentError.new(e.message)
         end
 
--- a/lib/hanami/entity/schema.rb
+++ b/lib/hanami/entity/schema.rb
@@ -93,11 +93,22 @@ module Hanami
         #
         # @since 0.7.0
         class Dsl
+          # @since x.x.x
+          # @api private
+          TYPES = [:schema, :strict, :weak, :permissive, :strict_with_defaults, :symbolized].freeze
+
+          # @since x.x.x
+          # @api private
+          DEFAULT_TYPE = TYPES.first
+
           # @since 0.7.0
           # @api private
-          def self.build(&blk)
+          def self.build(type, &blk)
+            type ||= DEFAULT_TYPE
+            raise Hanami::Model::Error.new("Unknown schema type: `#{type.inspect}'") unless TYPES.include?(type)
+
             attributes = new(&blk).to_h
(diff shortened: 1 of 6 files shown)


mereghost (approved):
LGTM. Awesome changes. We will need some nice documentation about this on the advanced guides =)

lib/hanami/entity/schema.rb
```diff
@@ -93,11 +93,22 @@ class Definition
         #
         # @since 0.7.0
         class Dsl
+          # @since x.x.x
+          # @api private
+          TYPES = [:schema, :strict, :weak, :permissive, :strict_with_defaults, :symbolized].freeze
```
cllns: I'm a little wary of copying these values from `dry-types`. I'm not sure there's a better solution though. If they add some new contstructor_type, it'd be nice to support it automatically.

`Dry::Types::Hash.public_instance_methods.first(6)` returns this same thing, but in a different order. This depends on their ordering of methods, which would be quite brittle, unfortunately :/

cllns (approved):
👍

jodosha:
@mereghost Yes, I referenced `dry-types` guides but we need our own.

PR #434, opened 2017-08-10
Wrap association exceptions

mereghost:
All the association code did not handle exceptions at all. The result was that we were receiving `ROM::SQL` exceptions instead of `Hanami::Model` ones.

This resulted in a very confused developer (myself) and some frustration, which leads us to this PR.

In all operations that might raise an exception, I added a `rescue` clause that wraps the error on a `Hanami::Model::Error`.

As, for the first time, I was working with the full association code, I made some changes to keep them consistent with each other.

Where the API changed I left an alias to the method old name just in case (and besides, it makes sense).

### DIFF
--- a/lib/hanami/model/associations/has_many.rb
+++ b/lib/hanami/model/associations/has_many.rb
@@ -62,6 +62,8 @@ module Hanami
         def add(data)
           command(:create, relation(target), use: [:timestamps])
             .call(associate(data))
+        rescue => e
+          raise Hanami::Model::Error.for(e)
         end
 
         # @since 0.7.0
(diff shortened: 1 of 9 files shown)


mereghost:
And mysql is being whiny. Checking.

cllns:
Could it be helpful to add `Hanami::Model::HasOneError` and `Hanami::Model::HasOneError` classes, for additional insight? (It might not, I'm not sure)

mereghost:
@cllns this patch just makes the association methods to behave like the repository ones. Almost all the exceptions raised would be from the database side.

The only one that would be nice to add would be one that deals with missing `id`s, but still not totally sold on that.

Good news is that I'm actively using most of the association code, so I get wind of any other nice stuff to have.

jodosha (approved):

PR #441, opened 2017-09-17
Improve connection string postgresql

AlfonsoUceda:
This PR implements https://github.com/hanami/hanami/issues/805

Check this link: https://www.postgresql.org/docs/9.6/static/libpq-connect.html#AEN45575

Tested in a fresh application with a connection string like the following:
```
posgresql:///foo_development?host=localhost&port=1234&user=username&password=password
```

cc @olistik

### DIFF
--- a/lib/hanami/model/sql/consoles/postgresql.rb
+++ b/lib/hanami/model/sql/consoles/postgresql.rb
@@ -36,7 +36,7 @@ module Hanami
           # @since 0.7.0
           # @api private
           def host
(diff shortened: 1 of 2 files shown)


mereghost (approved):
LGTM

AlfonsoUceda:
For the records, this change fix when using postgresql and jdbc driver.

AlfonsoUceda:
https://github.com/hanami/model/issues/452

PR #443, opened 2017-09-27
Repository#assoc scope delegation

mereghost:
Both `has_many` and `many_to_many` associations had this glaring issue where you could not filter any results of said association.

Let's say:
```
class AuthorRepository < Hanami::Repository
  associations do
    has_many :books
  end

  # works
  def books_for(author)
    assoc(:books, author)
  end

 # works
  def books_titled(author, title)
    assoc(:books, author).where(title: title)
  end 

  # did not work
  def last_books_published_for(author)
   assoc(:books, author).order(:published_at).reverse.limit(5)
  end
end
```
This PR fixes this by delegating a few methods back at `#scope` and returning a new association object.

### DIFF
--- a/lib/hanami/model/associations/has_many.rb
+++ b/lib/hanami/model/associations/has_many.rb
@@ -104,7 +104,7 @@ module Hanami
 
         private
 
-        def method_missing(meth, args)
+        def method_missing(meth, *args)
           whitelisted_methods = %i[where order limit reverse]
           return super unless whitelisted_methods.member?(meth) && scope.respond_to?(meth)
           __new__(scope.public_send(meth, args))
--- a/lib/hanami/model/associations/has_many.rb
+++ b/lib/hanami/model/associations/has_many.rb
@@ -96,11 +96,11 @@ module Hanami
           scope.to_a
         end
 
-        # @since 0.7.0
-        # @api private
-        def where(condition)
-          __new__(scope.where(condition))
-        end
+        # # @since 0.7.0
+        # # @api private
+        # def where(condition)
+        #   __new__(scope.where(condition))
+        # end
 
         # @since 0.7.0
         # @api private
@@ -109,6 +109,16 @@ module Hanami
         end
 
         private
+
+        def method_missing(meth, args)
+          whitelisted_methods = %i(where order limit)
+          return super unless whitelisted_methods.member?(meth) && scope.respond_to?(meth)
+          __new__(scope.public_send(meth, args))
+        end
+
+        def respond_to_missing?(meth, include_all)
+          scope.respond_to?(meth, include_all)
+        end
 
         # @since 0.7.0
         # @api private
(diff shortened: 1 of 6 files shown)


spec/integration/hanami/model/associations/has_many_spec.rb
```diff
@@ -173,4 +173,25 @@
...
     it '#remove'
   end
+
+  context 'delegates scope manipulation to #scope' do
+    let(:author) { authors.create_with_books(name: 'Luis de Camões', books: [{ title: 'Os Lusíadas' }]) }
+    let(:book_assoc) { authors.books_for(id: author.id) }
+
+    it 'responds to #where' do
+      expect(book_assoc).to respond_to :where
+      expect { book_assoc.where(title: 'Os Lusíadas') }.to_not raise_error
```
AlfonsoUceda: do you know what error can be raised? maybe we can be more specific.

lib/hanami/model/associations/has_many.rb
```diff
@@ -98,18 +98,22 @@ def to_a
...
-
-        # @since 0.7.0
-        # @api private
         def count
           scope.count
         end
 
         private
 
+        def method_missing(meth, args)
```
jodosha: This should be `*args`

lib/hanami/model/associations/many_to_many.rb
```diff
@@ -99,6 +95,16 @@ def remove(target_id)
 
         private
 
+        def method_missing(meth, args)
```
jodosha: This should be `*args`

mereghost:
@jodosha about the `belongs_to` bit IMO a many-to-one association, the whitelisted methods (limit, order, where etc) make little or no sense.

For the tests, I'm open to suggestions. I do agree that a more specific test would be nice, but these check if the object respond to the delegated methods. =/

jodosha:
@mereghost What's the observable behavior from the outside? That should be the subject of our tests.

One practical example: `AuthorRepository#last_books_published_for` assert that they are returned in the expected order.

mereghost:
I'll add an example with that case. My fear would be that I would end testing `rom` itself if I went down this route. =)

mereghost:
@jodosha Ok, mysql is acting up again and I have no clue why (except that it's mysql, that is. =p).
 
All in all, it seems more like an upstream issue than something we can fix: the error revolves around having parentesis around the LIMIT value, or so it seems. I can change the tests to not use `#limit` but the issue would still be there.

Generated query for reference (works on sqlite and pg without troubles):
```sql
SELECT `id`, `author_id`, `title`, `on_sale`, `created_at`, `updated_at` FROM `books` WHERE (`author_id` = 18) ORDER BY (`created_at`) DESC LIMIT (2)
```

mereghost:
I'm closing this for now as we plan our next steps. I'll get back at it later.

PR #448, opened 2017-10-03
createdb removed

micdahl:
Removed dependency to createdb and dropdb for Postgres, according to [this issue](https://github.com/hanami/model/issues/442).

### DIFF
--- a/lib/hanami/model/migrator/postgres_adapter.rb
+++ b/lib/hanami/model/migrator/postgres_adapter.rb
@@ -33,7 +33,16 @@ module Hanami
         def create
           set_environment_variables
 
-          call_db_command('createdb')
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@micdahl thanks for this PR, I've changed base to `develop` branch because when add new features we should target this branch, can you fix conflicts? you need rebase develop in your branch.

Checking the PR, you have old code but anyway you have removed `set_environment_variables` call and I think we need that, if you don't need for remote databases maybe we can avoid it, but we need it for local databases.

jodosha:
@micdahl Thanks for this PR, but CI build doesn't pass with these changes. Can you please check it? Thanks.

jodosha:
@micdahl I'm closing this as stale. Please feel free to reopen, when you'll have time to deal with this.

PR #453, opened 2017-10-11
Make it compatible with dry-types 0.12.1

jodosha:
`dry-types` `v0.12.1` introduced a breaking change: it doesn't longer let `ArgumentError` to be raised when _composite_ or _sum_ types are trying to coerce the input.

While I agree with that choice, that is a breaking change for our public API, so we had to fix it for now, and then go for an exception-less approach in 2.0.

---

Ref https://github.com/dry-rb/dry-types/commit/f535775b7c643b27a80de6690f92ec495cde93d2


jodosha:
Closing in favor of #454

PR #454, opened 2017-10-11
Depend on dry-types ~> 0.11.0

jodosha:
According to @flash-gordon, `rom-sql` isn't compatible with `dry-types` `0.12`, we should stay on `0.11`.


jodosha:
This fixes the build for `master` https://travis-ci.org/hanami/model/builds/286509924

PR #455, opened 2017-10-11
Set minimal rom-sql version

flash-gordon:
I made `rom-sql` 1.x compatible with `dry-types` 0.11 in https://github.com/rom-rb/rom-sql/pull/250. Dependency for `dry-types` is already set in master (https://github.com/hanami/model/pull/454), here I set the rom-sql's version and run CI check, once it's green I'll push 1.3.5 to rubygems.

### DIFF
--- a/hanami-model.gemspec
+++ b/hanami-model.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.3.0'
 
   spec.add_runtime_dependency 'hanami-utils',    '~> 1.0'
-  spec.add_runtime_dependency 'rom-sql',         '~> 1.3'
+  spec.add_runtime_dependency 'rom-sql',         '~> 1.3', '>= 1.3.5'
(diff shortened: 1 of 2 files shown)


jodosha:
@flash-gordon Thank you very much. In my understanding `rom-sql 1.3.5` should go with `dry-types 0.11` series. On the other hand, `rom-sql 1.3.4-` should go with `dry-types 0.10`. Is that correct?

flash-gordon:
@jodosha yes, you’re right

jodosha:
@flash-gordon Given `hanami-model` `v1.0.3` released on yesterday targets `dry-types` `~> 0.11.0`, according to your versioning explanation, this PR should go in `master` (`1.0.x` series), so we can release a `v1.0.4`.

flash-gordon:
@jodosha sorry, I messed this up, gimme a sec to rebase it against master again :)

flash-gordon:
@jodosha ok, done!

PR #457, opened 2017-10-12
Allow entities to be used as types in manual schema

### LINKED ISSUE #410: Entity as a type does not work
I think that "entity type" in entity attributes is a good idea and I want to use something like this:

```ruby
class Post < Hanami::Entity
  attributes do
    attribute :id,         Types::Int
    attribute :created_at, Types::Time
    attribute :updated_at, Types::Time
    attribute :deleted_at, Types::Time
    attribute :name,       Types::String
    attribute :body,       Types::String
    attribute :comments,   Types::Collection(Comment)
    attribute :user,       User
  end
end
```

But this code does not work. For fix this issue I need to add a hack like this:
```ruby
class Post < Hanami::Entity
  attributes do
    attribute :id,         Types::Int
    attribute :created_at, Types::Time
    attribute :updated_at, Types::Time
    attribute :deleted_at, Types::Time
    attribute :name,       Types::String
    attribute :body,       Types::String
    attribute :comments,   Types::Collection(Comment)
    attribute :user,       Hash
  end

  def user
    User.new(super)
  end
end
```

I think we can create a new type (like `Type::Entity(User)`) or update entity for using this as a type.

WDYT?

/cc @hanami/core

jodosha:
Introduce an enhancement for manual schema: allow entities to be used as types.

```ruby
require "hanami/model"

  attributes do
    # ...
    attribute :owner, Types::Entity(User)
  end
end

account = Account.new(owner: User.new(name: "Luca"))
account.owner.class # => User
account.owner.name  # => "Luca"

account = Account.new(owner: { name: "MG" })
account.owner.class # => User
account.owner.name  # => "MG"
```

---

Closes #410

### DIFF
--- a/lib/hanami/model/types.rb
+++ b/lib/hanami/model/types.rb
@@ -17,15 +17,67 @@ module Hanami
       # Class level interface
       #
       # @since 0.7.0
+      # rubocop:disable Naming/MethodName
       module ClassMethods
+        # Define an entity of the given type
+        #
+        # @param type [Hanami::Entity] an entity
+        #
+        # @since 1.1.0
+        #
+        # @example
+        #   require "hanami/model"
+        #
+        #   class Account < Hanami::Entity
+        #     attributes do
+        #       # ...
+        #       attribute :owner, Types::Entity(User)
+        #     end
+        #   end
+        #
+        #   account = Account.new(owner: User.new(name: "Luca"))
+        #   account.owner.class # => User
+        #   account.owner.name  # => "Luca"
+        #
+        #   account = Account.new(owner: { name: "MG" })
+        #   account.owner.class # => User
+        #   account.owner.name  # => "MG"
+        def Entity(type)
+          type = Schema::CoercibleType.new(type) unless type.is_a?(Dry::Types::Definition)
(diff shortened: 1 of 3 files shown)


lib/hanami/model/types.rb
```diff
@@ -17,15 +17,67 @@ def self.included(mod)
...
+        #   account = Account.new(owner: User.new(name: "Luca"))
+        #   account.owner.class # => User
+        #   account.owner.name  # => "Luca"
+        #
+        #   account = Account.new(owner: { name: "MG" })
+        #   account.owner.class # => User
+        #   account.owner.name  # => "MG"
+        def Entity(type)
+          type = Schema::CoercibleType.new(type) unless type.is_a?(Dry::Types::Definition)
+          type
```
davydovanton: WDYT about ternary operator: `type.is_a?(Dry::Types::Definition) ? type : Schema::CoercibleType.new(type)`?

davydovanton (approved):
great work 👍

AlfonsoUceda (approved):
@jodosha It works great!

PR #459, opened 2017-10-14
Fix repository/relation select

jodosha:
Allow relations in repositories to select a subset of columns

---

Fixes https://github.com/hanami/model/issues/451
Ref https://github.com/rom-rb/rom-repository/issues/91
Ref https://github.com/rom-rb/rom/commit/4c3b49fbea053bde3695147f8c2828239a40fac4


AlfonsoUceda (approved):
:+1:

PR #462, opened 2017-10-23
Associations to deep serialize input to `::Hash`

jodosha:
Let associations to deep serialize input to `::Hash`.

---

Closes https://github.com/hanami/controller/issues/235
Ref https://github.com/hanami/utils/pull/238

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ unless ENV['TRAVIS']
   gem 'yard',   require: false
 end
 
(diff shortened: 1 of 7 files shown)


AlfonsoUceda (approved):
LGTM but one thing, isn't there  a spec for many to many relationship?

jodosha:
@AlfonsoUceda because it already used to serialize input via `map(&:to_h)`. If you remove `serialize(data)` the spec fail.

PR #465, opened 2017-11-17
Fix typos in spec (intance -> instance)

autopp:
I found and fixed some typos in spec files. (intance -> instance)
I hope this helps.


autopp:
@mereghost @davydovanton Thanks for review and merge!

PR #468, opened 2017-11-27
adds find_by method to repository class

ippachi:
I made some hanami application. and I often defined find_by method like rails's one.
so I think it will be more convenient if hanami have find_by method.
What do you think?

### DIFF
--- a/lib/hanami/repository.rb
+++ b/lib/hanami/repository.rb
@@ -413,6 +413,21 @@ module Hanami
       raise Hanami::Model::Error.for(e)
     end
 
+    # Find by any conditions
+    #
+    # @return [Hanami::Entity,NilClass] one entity, if any
+    #
+    # @since 1.1.1
+    #
+    # @example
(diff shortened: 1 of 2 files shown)


jodosha:
@ippachi Hi and thanks for this PR, but it doesn't fit the current design of repository.

The role of the repository is to abstract low level queries from the outside world. For this reason, by design it's not possible to do:

```ruby
repository = CertificateRepository.new
repository.where(state: "issued").where(provider: "letsencrypt")
```

Instead, you are forced to pick a business name for your finder:

```ruby
repository = CertificateRepository.new
repository.letsencrypt_issued
```

Similarly, a `#find_by` method would introduce the same problem again:

```ruby
repository = CertificateRepository.new
repository.find_by(state: "issued", provider: "letsencrypt")
```

Thanks for understanding.

ippachi:
Ok. I understand. Thank you for explanation.

PR #471, opened 2018-01-25
Overwriting the `Repository#command` method

mereghost:
This PR eases end user pain when defining new commands.

Before she had to pass a bunch of options that she had no idea even where there or waht the Hanami defaults are, so we take care of that.

```ruby
  def create_many(bunch_of_data)
    command(create: :users, result: :many).call(bunch_of_data)
  end

  def ye_old_way_of_creating_many(bundle_of_data)
    command(create: :users, 
      mapper: :entity, 
      result: :many, use: [:schema, :mapping, :timestamps]).call(bundle_of_data)
  end
end
``` 

Closes hanami/hanami#856

### DIFF
--- a/spec/integration/hanami/model/repository/command_spec.rb
+++ b/spec/integration/hanami/model/repository/command_spec.rb
@@ -1,0 +1,27 @@
+RSpec.describe 'Customized commands' do
+  subject(:authors) { AuthorRepository.new }
+
+  let(:data) do
+    [{name: 'Arthur C. Clarke'}, {name: 'Phillip K. Dick'}]
+  end
+
+  context 'the mapper' do
+    it 'is enabled by default' do
+      result = authors.create_many(data)
+      expect(result).to be_an Array
+      expect(result).to all(be_an(Author))
+    end
+
+    it 'can be explictly turned off' do
+      result = authors.create_many(data, opts: { mapper: nil })
+      expect(result).to all(be_an(ROM::Struct))
+    end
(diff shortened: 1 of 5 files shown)


mereghost:
I tried to isolate the whole command deal, but in the end I either ended up with a mess, or had to reimplement half of `ROM::Repository::Root` by hand.

And still had some troubles dealing with custom commands like upsert.

spec/integration/hanami/model/repository/command_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  let(:data) do
+    [{name: 'Arthur C. Clarke'}, {name: 'Phillip K. Dick'}]
+  end
+
+  context 'the mapper' do
+    it 'is enabled by default' do
+      result = authors.create_many(data)
+      expect(result).to be_an Array
+      expect(result).to all(be_an(Author))
```
jodosha: TIL `all`. Cool! 😉

mereghost:
@jodosha I did some digging around using `upsert`. 

From what I could understand the `ROM::SQL::Commands::Postgres::Upsert` does not seem to get registered on the `ROM::Container`. 

So when we pass `:upsert` as an argument to `command` it's type (`ROM::SQL::Commands::Create`) is never found which in the end triggers the error.

The `upsert` method that is available on the relation only accepts the options for the conflict resolution and is used internally by the `Upsert` command.

PR #482, opened 2018-04-19
Use dry-inflector

### LINKED ISSUE #481: Use `dry-inflector`
Association names are generated by pluralizing/singularizing target entities/relations.
Internally these inflections are handled by `Utils::String.pluralize` and `.singularize`. These two methods will be removed in Hanami 2.0, alongside with `Utils::Inflector`.

For this reason, starting from 2.0, `hanami-model` should use `dry-inflector`. The main difference between `Utils::Inflector` and `dry-inflector` is the API: the first uses class level methods, whereas the second is an object, so it uses instance methods.

  – [ ] Let `Hanami::Model::Configuration` to accept a new dependency: `inflector`. See the snipped below
  - [ ] Replace `Utils::String.pluralize` and `.singularize` to use that internal `@inflector` instead

```ruby
def initialize(..., inflector: Dry::Inflector.new)
  # ...
  @inflector = inflector
end
```

---

Ref https://trello.com/c/nJhd5awo

artofhuman:
Closes #481

### DIFF
--- a/hanami-model.gemspec
+++ b/hanami-model.gemspec
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "rom-sql",         "~> 1.3", ">= 1.3.5"
   spec.add_runtime_dependency "rom-repository",  "~> 1.4"
   spec.add_runtime_dependency "dry-types",       "~> 0.11.0"
+  spec.add_runtime_dependency "dry-inflector",   "~> 0.1"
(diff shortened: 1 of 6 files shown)


artofhuman:
@jodosha thanks for review, i`ll try to implemet

jodosha:
@artofhuman Please let me know in case you'll need help. Thanks. 😃

artofhuman:
@jodosha ok ;)

cllns:
Already merged but looks fine to me too ;)

PR #509, opened 2018-12-05
Revert unintended change to has_many

mereghost:
IIRC when I was making sure that using rom commands like create(relation, result: :many) I needed to override the command method from rom.
During the exploration and testing I had to change the call to it inside has_many so the arguments were passed correctly.
And guess what? After I solved the issue I completely forgot about the change, since everything worked. >.<

### DIFF
--- a/lib/hanami/model/associations/has_many.rb
+++ b/lib/hanami/model/associations/has_many.rb
@@ -113,7 +113,7 @@ module Hanami
         # @since 0.7.0
         # @api private
         def command(target, relation, options = {})
-          repository.command(target => relation, **options)
+          repository.command(target, relation, options)
         end
 
         # @since 0.7.0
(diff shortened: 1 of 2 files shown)


mereghost:
I thought about it but could not figure a way, because we would need to somehow intercept the calls to Sequel way down the line. >.<

mereghost:
Sure, @melvrickgoh :) And thanks for helping us with this.

melvrickgoh:
hey @mereghost , apologies for the delay. Could finally dedicate some effort to do this. We've tested the fix including Hanami 1.3 as the control set and can verify that this branch is good.

I've attached our testing and results here for your reference

https://docs.google.com/document/d/1H0y9yRTAIFeDZ5g5WL4mvAcjUHbcPJ7RFKkMtCbmIKQ/edit?usp=sharing

We've also tested the branch by having this branch of hanami-model utilised in our company's staging environment which has integration specs testing various functionalities (hence a plethora of repository retrieval and storage functions) for about 3 days. 

No issues were reported, so we're confident that this is good for release too. 

Please feel free to let me know if you need more information.

melvrickgoh:
hello @mereghost @jodosha, do you think you could help us to check out the results and see if it meets what you're looking for? we've also run this for an extended period in production and it's working well for us. (~ 2 weeks, ~ -0.5M records)

mereghost:
@melvrickgoh Sorry for the long delayed response. It seems to match my expectations. =)

@jodosha mind if I merge this?

PR #534, opened 2019-05-30
ROM 5

jodosha:
## Goals

Upgrade to `rom` 5.0 and `dry-types` 1.0. `Hanami::Entity` is now powered by `dry-struct`.

### Entities

  * Dynamic classes for nested entities:

```ruby
class PageVisit < Hanami::Entity
  attribute :id, Types::Strict::Integer
  attribute :page_info do
    attribute :name, Types::Coercible::String
  end
end

entity = PageVisit.new(id: 1, page_info: { name: "index" })
entity.id # => 1
entity.page_info.class # => PageVisit::PageInfo – DYNAMIC CLASS
entity.page_info.name # => "index"
```

### Repositories

  * Removed repository's `aggregate` in favor of `Relation#combine`:

```ruby
class BookRepository < Hanami::Repository[:books]
  def find_with_categories(id)
    # old syntax
    aggregate(:categories).where(id: id).map_to(Book).one

    # new syntax
    books.combine(:categories).where(id: id).map_to(Book).one
  end
end
```

### Entities

  * Initializing an entity with wrong data raises `Hanami::Model::Error`, instead of `TypeError` or `ArgumentError`. Examples of wrong data: trying to pass a value for an attribute that fails a coercion, or omitting a required attribute for an entity with _strict_ schema.

  * Renamed `Types::Int` into `Types::Integer`.

  * Removed `Hanami::Entity.attributes` as a wrapper to define attributes:

```ruby
# old syntax
  attributes do
    attribute :id,   Types::Strict::Int
    attribute :name, Types::String
  end
end

# new syntax
  attribute :id,   Types::Strict::Int
  attribute :name, Types::String
end
```

  * _Strict_ policy is no longer passed to `Hanami::Entity.attributes(:strict) { }`, but with this new syntax:

```ruby
# old syntax
class Person < Hanami::Entity
  attributes :strict do
    attribute :id,   Types::Strict::Int
    attribute :name, Types::Strict::String
  end
end

# new syntax
class Person < Hanami::Entity[:strict]
  attribute :id,   Types::Strict::Integer
  attribute :name, Types::Strict::String
end
```

  * _Schemaless_ policy is defined with this new syntax:

```ruby
class Person < Hanami::Entity[:struct]
end
```

  * Removed the following schema policies: `nil`, `:schema`, `:weak`, `:permissive`, `:strict_with_defaults`, `:symbolized`. Allowed policies are: default, `:strict`, and `:struct`. Default policy doesn't require any manual definition of the schema **if** using SQL adapter, otherwise you must define attributes manually. This was already the default in 1.x.


gmanley:
@jodosha Thank you for your work! I do have one question: Why the change from `aggregate` to `combine`? Could you explain the rational? I imagine it's gonna be a change that affects a lot of apps.

jodosha:
@gmanley It's a ROM 5 change.

PR #556, opened 2019-10-11
Target Updated Default Travis CI Build Environment

landongrindheim:
Certain (Postgres-specific) specs are expecting syntax specific to newer versions of PostgreSQL (10 and up). hanami/model is currently relying on Travis CI. In April this year, [Travis CI updated its default build environment to Ubuntu Xenial 16.04](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment#1-services-support). Xenial allows Travis CI to run PostgreSQL 10, but it uses a pre-10 version out-of-the-box.

An example of a failure that occurred when running the PostgreSQL 10+ dependent code on a pre-10 version:

```bash
  1) Hanami::Model::Migrator PostgreSQL apply dumps database schema.sql
     Failure/Error:
       expect(actual).to include <<~SQL
         CREATE SEQUENCE public.reviews_id_seq
             AS integer
             START WITH 1
             INCREMENT BY 1
             NO MINVALUE
             NO MAXVALUE
             CACHE 1;
       SQL
```

I also found that MySQL was not running, resulting in builds failing when attempting to use the database. Per previously-linked blog post, "Services like MySQL or PostgreSQL are not started by default."

This pull request configures hanami/model's CI build to run in the updated Travis CI environment. It also removes no-longer-needed workarounds that were in place to run tests using the old Travis CI environment.

I see the value in not squashing these commits as I documented the reasons for these changes in database-specific commit messages. I will understand if that's not in line with the organization's standards.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ env:
   - DB=postgresql
   - DB=mysql
 
+services:
+  - mysql
+
 addons:
   postgresql: '10'
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
   - DB=mysql
 
 addons:
-  postgresql: '9.4'
+  postgresql: '10'
 
 matrix:
   allow_failures:
--- a/spec/integration/hanami/model/migration/mysql.rb
+++ b/spec/integration/hanami/model/migration/mysql.rb
@@ -392,8 +392,7 @@ RSpec.shared_examples 'migration_integration_mysql' do
       expect(options.fetch(:allow_null)).to eq(false)
 
       expected = Platform.match do
-        ci(:travis) { '0' }
-        default     { nil }
+        default { nil }
       end
 
       expect(options.fetch(:default)).to eq(expected)
@@ -409,8 +408,7 @@ RSpec.shared_examples 'migration_integration_mysql' do
       expect(options.fetch(:allow_null)).to eq(false)
 
       expected = Platform.match do
-        ci(:travis) { '0' }
-        default     { nil }
+        default { nil }
       end
 
       expect(options.fetch(:default)).to eq(expected)
--- a/spec/integration/hanami/model/migration/postgresql.rb
+++ b/spec/integration/hanami/model/migration/postgresql.rb
@@ -420,8 +420,7 @@ RSpec.shared_examples 'migration_integration_postgresql' do
       expect(options.fetch(:allow_null)).to eq(true)
 
       expected = Platform.match do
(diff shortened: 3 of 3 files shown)


landongrindheim:
@jodosha I did see that failure. Travis's support for Postgres versions below 10 seems to be a lot better than current versions. I'll be trying a few things to resolve that error and will ping you when it's done 🙂

landongrindheim:
@davydovanton This isn't ready yet. Travis is a bit behind the times and doesn't support Postgres versions greater than 9.4 (maybe 9.6) out of the box. This will require some additional setup, which should be made specific to the Postgres builds (i.e. probably not a `before_script`).

I'm considering reaching out to someone at Travis early next week to see if there's an established convention for dealing with this.

An alternative would be to have hanami/model join other Hanami repositories in using DroneCI.

davydovanton:
Hmmm, thanks for update. I also know that hanami uses circleCI for ci. Could you check your idea with it?

landongrindheim:
The change I've been trying to make is only applicable to TravisCI. As @davydovanton pointed out, configuration is present for CircleCI. DroneCI configuration is also present. Is there a reason for using one over the others?

The snag I've hit here is specific to TravisCI being behind the times and not doing a great job supporting newer versions of Postgres. I believe it can be overcome without too much effort, but I think it's a good idea to ask whether it's worth doing.

@mereghost @jodosha Do you have any thoughts?


It's also worth noting that this is one of two sets of failures in CI, the other being specific to MySQL. I've not tried addressing that yet.

landongrindheim:
@davydovanton @jodosha I tried a lot of different things before discovering that Travis CI had updated its build environment around six months ago. I've rebased changes into two commits based on the failures they were related to.

The approach changed significantly, but the result was better than I was expecting (only fixing Postgres builds).

This is ready for review 😄

landongrindheim:
@jodosha Just a friendly ping 🙂 Would you consider merging this pull request?

landongrindheim:
@jodosha This change only fixes CI. It doesn't alter library behavior.

Summary: `hanami/model` relied on a workaround in the test env to run against Travis's older PostgreSQL version. Travis updated their default CI environment, making the PostgreSQL workaround obsolete. Travis's default environment does not run MySQL by default, causing all MySQL builds to fail. Travis configuration has been updated to make the builds accurate.

If this is something you're not interested in, please let me know so I can close it and clear it from my GitHub dashboard.

jodosha:
@landongrindheim Thanks for this PR 👏

PR #563, opened 2019-12-20
Update dry-types 

mereghost:
This makes model alpha releasable. =)

We had an issue with dry-types `1.2` removing `.meta` from structs, which Nikita promptly fixed.

### DIFF
--- a/hanami-model.gemspec
+++ b/hanami-model.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "rom",             "~> 5.1"
   spec.add_runtime_dependency "rom-repository",  "~> 5.1"
   spec.add_runtime_dependency "rom-sql",         "~> 3.0"
-  spec.add_runtime_dependency "dry-types",       "~> 1.1"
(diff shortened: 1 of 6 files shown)


hanami-model.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "rom",             "~> 5.1"
   spec.add_runtime_dependency "rom-repository",  "~> 5.1"
   spec.add_runtime_dependency "rom-sql",         "~> 3.0"
-  spec.add_runtime_dependency "dry-types",       "~> 1.1"
+  spec.add_runtime_dependency "dry-types",       "~> 1.2"
```
davydovanton: WDYT if we set `1.3` version instead `1.2`? in this case, we'll able to use latest dry-system and I'll  have an ability to test my library for dry-system dependency graph 🤔

mereghost: IIRC it wasn't available at the time I made the changes. I'll give it a try and see how it goes.

mereghost: Okay, updating to 1.3 introduces new issues. I'll merge this, and open a new one to 1.3 otherwise it will get confusing.

PR #619, opened 2021-02-17
Fix postgresql in GitHub Actions

adam12:
The CI failing silently for postgresql came down to two issues:

1. The password isn't being set for `createdb` and `dropdb` correctly, causing the first `createdb` to sit forever. I've forced `PGPASSWORD` manually in `ci.yml` but this might be something we need to fix elsewhere (the postgres adapter maybe - see [this comment](https://github.com/hanami/model/issues/595#issuecomment-780735067)).
2. Depending on the rspec seed, the value of `PGPASSWORD` is being destroyed and never being reset during a spec run, causing further issues later. I've added an around hook as a temporary fix but we probably want to improve either the specs (ie. not call `ENV.delete`) or perhaps not modify `ENV` but have a `Hash` for `ENV` when calling `open3`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ name: ci
       - "Rakefile"
       - "Gemfile"
       - ".rubocop.yml"
+      - "script/ci"
   pull_request:
     branches:
       - master
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,7 @@ jobs:
           ruby-version: ${{matrix.ruby}}
       - name: Run all tests
         env:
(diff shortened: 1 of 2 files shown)


jodosha (commented):
@adam12 thanks for taking care of this. The commits in this PR don't have a green tick, nor I can find the build in the [list](https://github.com/hanami/model/actions).

Where can I verify that the PR fixed CI? Thanks.

adam12:
@jodosha This should be good now. I didn't want to waste Hanami CI minutes while debugging this so I had originally done the work in my own repo.

But I've pushed to `hanami/model` and it's now listed.

jodosha (approved):

PR #620, opened 2021-02-27
Ignore `schema_migrations` and `schema_info` errors

wenderjean:
Related to: https://github.com/hanami/hanami/issues/1087

### Changes

- Implement method `error` from STD Lib [Logger](https://github.com/ruby/logger/blob/master/lib/logger.rb#L546) in order to suppress the lines we don't want to show in the output
- Create a simple expectation to cover the new behavior

### Rational

The `Sequel` gem yields an error log for the absence of `schema_migrations` and/or `schema_info` table right before creating it and, for that reason, we can see those errors the first time we try to prepare/migrate the database even though it doesn't produce blockers.

### Inputs

The errors we are getting for each adapter are:

#### SQLite
```
SQLite3::SQLException: no such table: schema_migrations: SELECT NULL AS 'nil' FROM `schema_migrations` LIMIT 1
SQLite3::SQLException: no such table: schema_info: SELECT NULL AS 'nil' FROM `schema_info` LIMIT 1
```

#### Postgres
```
PG::UndefinedTable: ERROR:  relation "schema_migrations" does not exist
PG::UndefinedTable: ERROR:  relation "schema_info" does not exist
```

#### MySQL
```
Mysql2::Error: Table 'bookshelf_development.schema_migrations' doesn't exist: SELECT NULL AS `nil` FROM `schema_migrations` LIMIT 1
Mysql2::Error: Table 'bookshelf_development.schema_info' doesn't exist: SELECT NULL AS `nil` FROM `schema_migrations` LIMIT 1
```

### DIFF
--- a/lib/hanami/model/migrator/logger.rb
+++ b/lib/hanami/model/migrator/logger.rb
@@ -40,7 +40,7 @@ module Hanami
           super(nil, stream: stream, formatter: Formatter.new)
         end
 
-        # @since <version>
+        # @since x.x.x
         # @api public
         def error(progname = nil, &block)
           return true if _ignorable(progname)
@@ -50,7 +50,7 @@ module Hanami
 
         private
 
-        # @since <version>
+        # @since x.x.x
         # @api private
         def _ignorable(progname)
           IGNORABLE_PATTERNS.any? { |r| progname =~ r }
--- a/lib/hanami/model/migrator/logger.rb
+++ b/lib/hanami/model/migrator/logger.rb
@@ -10,6 +10,16 @@ module Hanami
       # @since 1.0.0
       # @api private
       class Logger < Hanami::Logger
+        # Messages patterns to identify errors related to both "schema_migrations" and "schema_info" absence tables.
+        #   1. SQLite
+        #   2. Postgres
+        #   3. MySQL
+        IGNORABLE_PATTERNS = [
+          /(?<=no such table: )(?:schema_migrations|schema_info)/,
+          /(?<=relation )(?:"schema_migrations"|"schema_info")(?= does not exist)/,
+          /(\.schema_migrations\'|\.schema_info\')(?= doesn\'t exist)/
+        ].freeze
+
         # Formatter for migrations logger
         #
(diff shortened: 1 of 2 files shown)


lib/hanami/model/migrator/logger.rb
```diff
@@ -29,6 +39,22 @@ def _format(hash)
         def initialize(stream)
           super(nil, stream: stream, formatter: Formatter.new)
         end
+
+        # @since <version>
```
wenderjean, line 43: I'm not sure how we usually feed this information.

adam12: Usually `# @since x.x.x`

wenderjean: Considering the last tag is v1.3.2 I suppose I should increment a patch version and put `since 1.3.3`?

adam12: Actually I believe it's literally `x.x.x`, and then backfilled at release.

lib/hanami/model/migrator/logger.rb
```diff
@@ -29,6 +39,22 @@ def _format(hash)
...
+
+          super(progname, &block)
+        end
+
+        private
+
+        # @since x.x.x
+        # @api private
+        def _ignorable(progname)
+          IGNORABLE_PATTERNS.any? { |r| progname =~ r }
```
adam12, line 56: Minor nit.

```suggestion
          IGNORABLE_PATTERNS.any? { |r| progname.match?(r) }
```

adam12 (commented):
Other than minor nit, looks good to me. I only tested the SQLite integration.

adam12:
@wenderjean is this ready to go?

wenderjean:
It is @adam12

adam12 (approved):
LGTM 👍

adam12:
Going to merge after we get a `v1.3.4` milestone added to the repo.

adam12:
Thanks for the patch @wenderjean !

PR #622, opened 2021-03-05
Use environment variables when interacting with PostgreSQL CLI tools

### LINKED ISSUE #595: db prepare raises Postgres createdb error
I encountered error from postgres version 13 in `createdb` command while trying to run `hanami db prepare` on clean app.

error trace shows problem in `path/hanami/model/migrator/postgres_adapter.rb:103`. 

Checking `command_with_credentials` method I see there's password option applied for database command:
```ruby
result << "--password=#{password}" unless Utils::Blank.blank?(password)
```
I thought it might be the case, that postgres version 13 dropped support of passing password value to this option, but it looks like there was never such possibility.

https://www.postgresql.org/docs/9.1/app-createdb.html

Please, correct me if I am wrong.

adam12:
Instead of using command line flags, environment variables are set for the execution of any PostgreSQL command.

The migrator spec builds a URL that excludes the password, but that URL is then used to build a connection, which is then used to run a PostgreSQL command. That command will not receive a password because the password in the connection string was absent.

Remove the `ci.yml` hack that forces `PGPASSWORD` to be present.

Closes #595

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,6 @@ jobs:
           ruby-version: ${{matrix.ruby}}
       - name: Run all tests
         env:
-          PGPASSWORD: root
           HANAMI_DATABASE_USERNAME: root
           HANAMI_DATABASE_PASSWORD: root
           HANAMI_DATABASE_HOST: 127.0.0.1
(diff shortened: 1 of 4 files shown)


wenderjean (approved):
Looks good, I performed successful tests creating/migrating a database using those versions below:

- [x] Postgresql 11
- [x] Postgresql 12
- [x] Postgresql 13

adam12:
> Looks good, I performed successful tests creating/migrating a database using those versions below:
> 

Much thanks!

